Issue #374, opened 2026-06-22
[Plugin] Marketplace install does not repair stale codestory-cli runtime on PATH


TheGreenCedar:
Parent: #38
Refs: #373, #372

Smoke finding from issue #373 on Windows after the 0.11.5 release.

Observed:
- Current GitHub release: v0.11.5, published 2026-06-22T07:08:11Z.
- Before repair, `where.exe codestory-cli` resolved `C:\Users\alber\AppData\Local\CodeStory\bin\codestory-cli.exe` and `codestory-cli --version` returned `codestory-cli 0.11.2`.
- `codex plugin add codestory@TheGreenCedar --json` succeeded and refreshed plugin package `codestory@TheGreenCedar` version `0.1.0` into `C:\Users\alber\.codex\plugins\cache\TheGreenCedar\codestory\0.1.0`.
- After that marketplace/plugin reinstall, `codestory-cli --version` still returned `codestory-cli 0.11.2` from the same PATH binary.
- The installed plugin MCP metadata launches direct CLI stdio: `codestory-cli serve --stdio --refresh none`; no Node wrapper was present.
- Three existing `codestory-cli.exe serve --stdio --refresh none` processes held the stale binary locked, so direct overwrite failed with `The process cannot access the file ... because it is being used by another process`.
- Installing the 0.11.5 release asset into an earlier PATH directory (`C:\Users\alber\.local\bin`) repaired new-process resolution: `where.exe codestory-cli` listed `.local\bin` first and `codestory-cli --version` returned `codestory-cli 0.11.5`.

Why this matters:
- Marketplace/plugin install can appear successful while the active/new MCP runtime still uses a stale CLI if PATH points at an older locked binary.
- Existing direct CLI stdio is the right runtime shape, but the install/update flow needs a clearer stale-runtime guard or bootstrap/update contract.

Scope:
- Make the CodeStory plugin/runtime setup path detect and report stale `codestory-cli` on PATH before MCP use, or document/automate a host-safe repair path that accounts for locked running stdio servers.
- Prefer the smallest CodeStory-owned surface: plugin package docs/skill, install helper, setup script, or static guard. Do not move marketplace catalog files into this repo.
- Preserve direct `codestory-cli serve --stdio --refresh none`; do not add a Node MCP wrapper.
- Preserve cross-platform intent. The observed repro is Windows, but the fix should not make macOS/Linux worse.

Acceptance:
[...]

Issue #373, opened 2026-06-22
[Dogfood] smoke 0.11.5 plugin install and runtime path


TheGreenCedar:
﻿Parent: #38
Refs: #357, #371

Smoke the released `0.11.5` CodeStory plugin/runtime path after the docs and setup promotion wave.

Scope:
- Treat `TheGreenCedar/AgentPluginMarketplace` as the external catalog owner; do not move marketplace files into this repo.
- Verify the CodeStory plugin package installs or updates through the marketplace/catalog flow to the latest released CLI, not a pinned stale binary.
- Verify the agent-facing runtime path uses the direct CodeStory CLI stdio/MCP surface, not a throwaway Node.js wrapper.
- Verify core agent tools are exposed well enough for grounding workflows: at minimum health/status, index/refresh, ground, search/context or equivalent read paths, and failure messages for missing sidecars.
- Record any dogfood friction: stale caches, locked binaries, platform assumptions, setup ambiguity, missing MCP tools, or sidecar readiness confusion.

Acceptance:
- Comment with exact environment, commands, installed plugin/runtime version, release asset used, and exposed tool list.
- If the smoke passes, state what was proven and what remains unproven.
- If it fails, open the smallest follow-up issue or PR-sized issue with the failing layer isolated.
- No source edits unless the failure is a small CodeStory-owned plugin/package/docs bug; any PR targets `dev/codestory-next`.

Must not break:
- Marketplace catalog ownership stays outside this repo.
- Install/update flow remains cross-platform in intent: Windows, macOS, and Linux.
- Docs must not tell humans to run CodeStory primarily as a manual CLI when the plugin/agent path is the surface being tested.

TheGreenCedar:
Coordinator started visible child lane. Pending worktree: `local:c8a34a2b-04a9-459e-a570-61eb01292756`. Target branch requested: `codex/issue-373-plugin-0.11.5-smoke`. PR target, if a CodeStory-owned fix is needed: `dev/codestory-next`.

TheGreenCedar:
Coordinator note from #372 docs audit: the child lane found `codestory-cli --version` on PATH still reports `0.11.2` while GitHub release `v0.11.5` is current. Please include this in the #373 install/runtime smoke: determine whether marketplace/plugin update resolves to 0.11.5 despite the stale PATH binary, or whether this is a remaining setup/runtime footgun.

TheGreenCedar:
[...]

Issue #38, opened 2026-06-18
[Saga] map code intelligence surfaces and first PR landing zones


TheGreenCedar:
Summary: Track the CodeStory product-grade intelligence saga across release, plugin packaging, sidecar reliability, docs quality, and MCP/plugin grounding lanes.

## Current State

Checkpoint: 2026-06-23 after PR #417, public release `v0.11.10`, and AgentPluginMarketplace PR #1.

- `main` is current at `70a204ce` (`release codestory 0.11.10`).
- Latest GitHub release is `v0.11.10`, published 2026-06-23T11:56:27Z.
- `dev/codestory-next` already contained the promoted stale-CLI and stdio sidecar fixes through `2debc4cf`; PR #417 promoted the release-sized slice to `main`.
- Open CodeStory PRs: none.
- Open CodeStory issues: none.
- Open AgentPluginMarketplace PRs/issues: none.
- Project #1 active rows: none outside completed items.
- Coordinator checkout: `C:\Users\alber\source\repos\codestory`, branch `main`, clean against `origin/main`.
- Active PATH resolves `codestory-cli` to `C:\Users\alber\.local\bin\codestory-cli.exe`; `codestory-cli --version` reports `codestory-cli 0.11.10`.
- Release asset proof: downloaded `codestory-cli-v0.11.10-windows-x64.zip` from GitHub release `v0.11.10`; SHA256 matched `c3b4799caad4e1885d97871347e1a5a03fcb5a716510cf4428907a3f300bb22d`; installed that binary into the first PATH location.
- Local Codex plugin cache proof: seeded `C:\Users\alber\.codex\plugins\cache\TheGreenCedar\codestory\0.11.10` with the full plugin wrapper; cached manifest reports version `0.11.10` and `.mcp.json` still invokes `codestory-cli` from PATH. Canonical repo plugin static tests pass 5/5.
- Marketplace propagation proof: TheGreenCedar/AgentPluginMarketplace PR #1 merged at `b47689e`; marketplace README now points CodeStory at `main/plugins/codestory` and `v0.11.10`; catalog JSON still parses and keeps the branchless git-subdir source.
[...]

PR #379, opened 2026-06-22
[codex] repair codestory runtime path

TheGreenCedar:
## Context

Closes #374
Refs #38, #373

#373 showed that refreshing the CodeStory plugin package did not guarantee the MCP runtime was current: `.mcp.json` launches `codestory-cli serve --stdio --refresh none` from the agent host `PATH`, and old Windows stdio server processes can keep the stale executable locked. PR #377 already fixed the separate plugin package version sync gap through #378; this PR handles the remaining runtime/PATH repair boundary.

```mermaid
flowchart LR
    Installer["install-codestory.ps1"] --> Pick["current cli selected"]
    Pick --> Path["put cli dir first on PATH"]
    Path --> Check["codestory-cli --version from PATH"]
    Check --> MCP["direct stdio MCP launch"]
    Check -->|stale| Fail["explain locked server and PATH order"]
```

## What Changed

- Updated `scripts/install-codestory.ps1` so the accepted or downloaded current CLI directory is moved to the front of user/process `PATH` when it lives under the managed install tree.
- Added a pre-launch PATH assertion that fails if `codestory-cli` still resolves to a stale binary, with repair text for locked `serve --stdio --refresh none` processes and PATH ordering.
- Kept the locked-default fallback, but clarified that the versioned fallback directory is used for new MCP launches.
- Added plugin README and grounding-skill guidance for the same locked-server/PATH-order repair path, plus static assertions so that guidance stays present.

## How To Review

- Check that `.mcp.json` still launches direct `codestory-cli serve --stdio --refresh none`; there is no Node wrapper and no marketplace catalog change.
- Review the installer path ordering logic and the explicit stale-PATH failure message.
- Confirm the docs describe the same repair path an agent or user needs before launching a fresh Codex thread.

## Verification

- `.\scripts\install-codestory.ps1 -SelfTest` -> pass. Covers locked default binary fallback, stale-first PATH failure, and repaired PATH ordering.
- `node --test plugins\codestory\tests\plugin-static.test.mjs` -> pass, 5 tests.
- `git diff --check` -> pass.

## Risk

The automatic release download path remains Windows x64 only, as before. macOS/Linux intent stays in the plugin skill/docs: verify the `codestory-cli` command from PATH before launch, keep the direct stdio runtime, and use source fallback when no release asset fits the host.

PR #380, opened 2026-06-22
Guard stale CodeStory plugin runtime

TheGreenCedar:
## Context

Issue #374 stayed open after #377/#378 because plugin package metadata now tracks the CLI version, but installed MCP runtime still launches `codestory-cli serve --stdio --refresh none` from the agent host `PATH`. A marketplace/plugin refresh can therefore leave new MCP launches on an older `codestory-cli` if PATH still resolves a stale or locked binary.

Closes #374
Refs #38, #373

## What Changed

- Added `plugins/codestory/scripts/check-runtime.mjs`, a dependency-free preflight that reads the installed plugin package version, resolves `codestory-cli` from the same `PATH` used by `.mcp.json`, runs `--version`, and fails if the resolved runtime does not match.
- Updated the plugin README and shipped `codestory-grounding` skill setup gate to run that preflight before a fresh MCP/plugin thread.
- Added repair guidance for PATH ordering, locked `codestory-cli serve --stdio --refresh none` processes, and Codex host/app restarts after PATH changes.
- Extended the plugin static test with a fake stale `codestory-cli` earlier on PATH so the guard and repair text are covered.

## How To Review

1. Read `plugins/codestory/scripts/check-runtime.mjs` first; it is the only new executable surface.
2. Check that `.mcp.json` still launches direct CLI stdio and no Node adapter was added.
3. Confirm the README and skill guidance point users/agents to the preflight before starting a fresh plugin thread.
4. Review the stale-runtime test in `plugins/codestory/tests/plugin-static.test.mjs`.

## Verification

- `node plugins\codestory\scripts\check-runtime.mjs` -> passed; resolved `C:\Users\alber\.local\bin\codestory-cli.exe` version `0.11.5`.
- `node --test plugins\codestory\tests\plugin-static.test.mjs` -> passed; 6 tests.
- `node --check plugins\codestory\scripts\check-runtime.mjs` -> passed.
- `git diff --check` -> passed.
- `codestory-cli doctor --project . --format json` -> read-only snapshot completed; this worktree is not indexed and reports `retrieval_manifest_missing`, so packet/search was not used as proof.

## Risk

- The preflight is an explicit command, not a wrapper around MCP launch. That preserves the direct runtime shape but depends on agents/users running the check before starting a fresh plugin thread.
- It compares against the plugin package version, which is now guarded to match `codestory-cli`; if future plugin compatibility intentionally spans multiple CLI versions, this check would need a version range.

### DIFF
--- a/plugins/codestory/README.md
+++ b/plugins/codestory/README.md
@@ -77,6 +77,20 @@ unavailable.
 Start a new Codex thread after installation or refresh. The installed package
 launches `codestory-cli serve --stdio --refresh none` from `PATH`.
 
+Before starting that fresh thread, run the runtime preflight from this plugin
+package:
+
+```console
+node plugins/codestory/scripts/check-runtime.mjs
+```
+
+It checks the `codestory-cli` resolved from `PATH` against the plugin package
+version. If it fails, install the matching release asset into a stable user bin
+directory, put that directory before older `codestory-cli` entries on `PATH`,
+stop existing `codestory-cli serve --stdio --refresh none` processes before
+replacing a locked binary, restart the Codex host/app if `PATH` changed, and
+then start a fresh thread.
+
 ### After install
 
 Open Codex in the repo you want to ground and ask the agent to check readiness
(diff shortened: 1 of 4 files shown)


TheGreenCedar:
Coordinator close: superseded by PR #379.

PR #379 already merged to `dev/codestory-next`, closed #374, and repaired the stale runtime/PATH boundary through the existing installer plus docs/static guards. This draft is now dirty against `dev/codestory-next`, still closes the already-closed #374, and adds a second executable preflight surface (`plugins/codestory/scripts/check-runtime.mjs`) that is not needed for the accepted fix.

Keeping the smaller merged path; no source from this draft is being promoted.

Issue #395, opened 2026-06-22
Classify missing CodeStory MCP namespace as plugin registration failure


TheGreenCedar:
## Context

A CodeStory dogfood session had the `codestory` plugin enabled and the skill loaded, but the `mcp__codestory` namespace was not discoverable in that thread. The installed package declared `.mcp.json` correctly, so treating CLI fallback as ordinary "MCP unavailable" hid the real setup failure.

## Acceptance Criteria

- CodeStory skill guidance tells agents to check for `mcp__codestory` tools or `codestory://status` when CodeStory is explicitly invoked.
- If the skill is visible but the MCP namespace/resource is missing, guidance calls it a plugin MCP registration failure.
- CLI fallback is documented as degraded evidence collection, not proof that the installed MCP surface is live.
- Plugin static checks prevent the guidance from regressing.

## Verification

- `node --test plugins\codestory\tests\plugin-static.test.mjs`
- `git diff --check`

TheGreenCedar:
Completed by merged PR #396 into `dev/codestory-next`.

Evidence:
- PR #396 merged at `e992ff5a`.
- The CodeStory skill now classifies skill-visible/MCP-missing as plugin MCP registration failure.
- README documents CLI as degraded evidence, not proof of live installed MCP.
- Plugin static test pins the wording.
- Verification on the PR: `node --test plugins\codestory\tests\plugin-static.test.mjs` passed 5/5, `git diff --check` passed, and delegated read-only review found no blockers.

GitHub did not auto-close this issue because the PR targeted `dev/codestory-next`, so closing manually with the merged PR evidence.

PR #396, opened 2026-06-22
Document CodeStory MCP registration failure

TheGreenCedar:
## Context

A dogfood session invoked `@CodeStory` with the plugin enabled, but the thread exposed the skill without the `mcp__codestory` namespace or `codestory://status` resource. The package manifest and `.mcp.json` were correct, so the failure should be named as plugin MCP registration failure instead of normalized as generic CLI fallback.

Closes #395

## What Changed

- Added a CodeStory skill setup-gate check for explicit `@CodeStory` use: if the skill is visible but MCP tools/resources are missing, report plugin MCP registration failure.
- Documented that CLI fallback can collect degraded health evidence but does not prove the installed MCP surface is live.
- Extended the plugin static test to keep the registration-failure wording in both the skill and README.

```mermaid
flowchart TD
    A["@CodeStory invoked"] --> B{"mcp__codestory or codestory://status exposed?"}
    B -->|yes| C["Use MCP readiness gate"]
    B -->|no| D["Report plugin MCP registration failure"]
    D --> E["Use CLI only as degraded fallback"]
```

## How To Review

- Read the setup-gate addition in `plugins/codestory/skills/codestory-grounding/SKILL.md`.
- Read the matching operator note in `plugins/codestory/README.md`.
- Confirm the static test asserts the new guidance without adding runtime scaffolding.

## Verification

- `node --test plugins\codestory\tests\plugin-static.test.mjs` passed, 5/5.
- `git diff --check` passed.

## Risk

This does not fix Codex host/tool-registry registration. It makes the CodeStory package guidance classify and report that failure truthfully when it happens.


TheGreenCedar:
Read-only delegated review result: no blocking findings.

Reviewer checked the three-file docs/static-test diff against the finding: skill visible but `mcp__codestory` / `codestory://status` missing must be classified as plugin MCP registration failure, with CLI only as degraded fallback.

Reviewer verification:
- `node --test plugins/codestory/tests/plugin-static.test.mjs` passed, 5/5.
- `git diff --check` passed.

Residual risk: this PR does not verify or fix the Codex host/plugin registry. It only makes the CodeStory package guidance report that host setup failure truthfully.

Issue #550, opened 2026-06-25
Split readiness truth by agent and local runtime lanes


TheGreenCedar:
Parent: #548
Related: #501, #470, #496

## Problem
CodeStory can currently report agent packet/search readiness while another status lane reports retrieval unavailable. Local repair/status paths can also expose or change sidecar drift after agent retrieval has been made ready. The states may be technically explainable, but the product contract is not clear enough for agents or maintainers.

## Acceptance Criteria
- `doctor`, status, and preflight report separate readiness lanes, including `agent_packet_search` and `local_default`.
- Each lane includes profile/run id, sidecar mode, degraded reason, and next command scoped to that lane.
- No output presents `ready` and `retrieval unavailable` as one undifferentiated truth.
- Local repair cannot silently invalidate agent sidecar manifests/readiness; if it changes another lane, status reports that explicitly.
- Regression covers agent full/local unavailable, local repair after agent ready, stale manifest/profile mismatch, and one-lane-ready/one-lane-blocked output.

## Non-Goals
- Does not change retrieval ranking.
- Does not add new sidecar storage.

TheGreenCedar:
Closed by #563, merged to dev/codestory-next as f7e5c0909168d9014253d0189c7542ae482fcc84. The PR splits local/default and agent packet/search readiness lanes, keeps compatibility fields, and makes missing agent run state explicit without collapsing to local profile.

PR #563, opened 2026-06-25
[codex] Split readiness lane reporting

TheGreenCedar:
## Context

Closes #550.

`doctor`, `agent preflight`, and `codestory://status` already had readiness verdicts, but the sidecar truth could still read as one collapsed state when agent packet/search and local/default sidecar readiness diverged.

```mermaid
flowchart LR
    S["sidecar status probes"] --> L["local_default lane"]
    S --> A["agent_packet_search lane"]
    L --> O["doctor / preflight / status JSON"]
    A --> O
```

## What Changed

- Added explicit readiness lane output for `local_default` and `agent_packet_search`.
- Included profile, run id, sidecar mode, degraded reason, status, and lane-scoped next command in lane output.
- Kept existing compatibility fields (`readiness`, `local_graph`, `full_retrieval`, top-level retrieval fields) intact.
- Added profile/run id to sidecar snapshots used by readiness verdicts.
- Covered doctor JSON, agent preflight JSON, stdio status JSON, and the profile-mismatch helper case.

## How To Review

- Inspect `crates/codestory-cli/src/main.rs` around `build_readiness_lanes_for_runtime` and `readiness_lane_status`.
- Confirm `crates/codestory-cli/src/stdio_transport.rs` includes `readiness_lanes` in `codestory://status`.
- Confirm the output structs in `crates/codestory-cli/src/args.rs` preserve existing fields and only add lane fields.

## Verification

- `cargo check -p codestory-cli`
- `cargo fmt`
- `cargo test -p codestory-cli readiness_lane_keeps_agent_full_separate_from_local_handoff_mismatch`
- `cargo test -p codestory-cli --test cli_golden_path doctor_next_commands_stop_at_retrieval_repair_when_sidecar_is_not_full`
- `cargo test -p codestory-cli --test cli_golden_path agent_preflight_reports_local_graph_when_retrieval_is_degraded`
- `cargo test -p codestory-cli --test stdio_protocol_contracts resources_read_status_reports_browser_readiness_and_next_calls`
- `git diff --check`

## Risk

- This is reporting-only. It does not change retrieval ranking, sidecar storage, or repair execution.
- The compatibility fields remain, so consumers can migrate to `readiness_lanes` without losing old fields.

### DIFF
--- a/crates/codestory-cli/src/args.rs
+++ b/crates/codestory-cli/src/args.rs
@@ -17,7 +17,7 @@ use codestory_contracts::api::{
     TrailCallerScope, TrailContextDto, TrailDirection, TrailMode,
 };
 use serde::{Deserialize, Serialize};
-use std::path::PathBuf;
+use std::{collections::BTreeMap, path::PathBuf};
 
 const INDEX_REFRESH_HELP: &str = "Index defaults to `auto`: it chooses `full` for an empty cache and `incremental` once the \
 cache already has indexed files.";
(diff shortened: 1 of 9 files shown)


TheGreenCedar:
Coordinator review gate: keep draft.

Independent review found a blocker in the lane contract: `agent_packet_search` currently uses the auto sidecar selector, which can fall back to the local/default profile when no agent run is present. That collapses the two lanes instead of making agent packet/search truth explicit.

Required before ready:
- `local_default` remains explicitly local/default.
- `agent_packet_search` explicitly represents the agent sidecar lane or a clear agent-run-missing/repair state.
- `agent_packet_search` must not silently report `profile=local` just because no agent run is active.
- Add focused regression coverage for the no-agent-run case plus the existing handoff mismatch behavior.

A fix lane is active; no merge until that lands and checks/review pass.

TheGreenCedar:
CI follow-up: linux-contracts failed in `resources_read_status_reports_browser_readiness_and_next_calls` because the stdio contract test asserted `run_id=agent-run-missing` even when CI provides an actual agent sidecar runtime.

The product contract should be:
- agent lane must report `profile=agent`, never collapse to `profile=local`;
- `run_id=agent-run-missing` only when there is no agent run/context;
- when an agent runtime exists, a concrete agent run id is valid and better evidence.

Fix lane is active to relax the stdio test accordingly while preserving the controlled no-agent regression coverage.

Issue #560, opened 2026-06-25
Add local graph wait-fresh repair for MCP startup


TheGreenCedar:
Parent: #554
Related: #496, #548, #549, #550, #551

## Problem
When the packaged MCP launcher sees a stale local graph, it currently fails open and hands the agent a repair command. That pushes the agent into command-by-command setup instead of letting the active CodeStory runtime perform one bounded local refresh.

## Acceptance Criteria
- Add or define local graph `ready/status --wait-fresh` semantics for a bounded local-only freshness wait/repair.
- MCP startup runs at most one bounded local graph refresh before launching `serve --stdio --refresh none`.
- Refresh uses existing full-vs-incremental `ensure_open(Auto)` behavior and existing indexing lock paths where possible.
- Failure or timeout keeps current fail-open diagnostic behavior with one clear degraded state.
- Does not start full packet/search retrieval sidecars.
- Tests cover stale local graph at MCP startup, successful local repair, timeout/failure, and no packet/search sidecar startup.

## Non-Goals
- Does not implement daemon/watch behavior.
- Does not add CodeStory to PATH.

TheGreenCedar:
Coordinator note after active MCP/server freshness spike:

This issue remains the right PR-sized owner for removing command-by-command local index setup at MCP startup. No duplicate issue is needed.

Recommended implementation shape:
- If the MCP launcher local navigation probe returns `repair_index`, run exactly one bounded `ready --goal local --repair --project <cwd> --format json`.
- Re-probe after repair; on success launch the existing `serve --stdio --refresh none` path.
- On timeout/failure, keep fail-open diagnostics with one clear degraded state.
- Keep this local-graph-only. Do not call `ready --goal agent --repair` and do not auto-build full packet/search sidecars.
- Keep PATH out of scope.

Companion status work belongs in #562: expose fresh/refreshing/failed/timed-out local refresh state separately from packet/search readiness.

TheGreenCedar:
Closed by #564, merged to dev/codestory-next as 0572715462c507e92f32630db99d28c19e276f61. The PR adds one bounded local-only MCP startup repair, preserves fail-open diagnostics, and keeps packet/search sidecar repair out of the repaired-local startup path.

PR #564, opened 2026-06-25
[codex] Repair local graph before MCP serve

TheGreenCedar:
## Context

Closes #560

Packaged MCP startup should not hand agents setup archaeology when only the local graph is stale. It can make one bounded local-only repair attempt, then either serve from a fresh graph or keep the existing diagnostic fail-open surface.

## What Changed
- Added one bounded local-only `ready --goal local --repair` attempt in the packaged MCP launcher when the initial local navigation probe is stale.
- Preserved fail-open MCP diagnostics on repair failure, invalid JSON, or timeout; diagnostics now include probe and repair command/setup evidence.
- Added plugin static coverage for successful local repair before serve, still-stale fail-open behavior, timeout fail-open behavior, and no agent sidecar repair during the local repair path.

## How To Review
- Start with `plugins/codestory/scripts/codestory-mcp.cjs` around `probeLocalNavigation`.
- Then review `plugins/codestory/tests/plugin-static.test.mjs` local navigation launcher tests.

## Verification
- `node --test --test-name-pattern "local navigation|repairs stale" plugins/codestory/tests/plugin-static.test.mjs`
- `node --test plugins/codestory/tests/plugin-static.test.mjs`
- `git diff --check`
- `.\target\debug\codestory-cli.exe ready --goal local --repair --project "C:\Users\alber\.codex\worktrees\2c54\codestory" --format json` -> local_navigation ready, 274/274 indexed, retrieval_manifest_missing unchanged.

## Risk
- Local repair timeout defaults to 120s; it is intentionally local-only and does not invoke `ready --goal agent` or packet/search sidecar startup.
- Cargo/Rust was not touched, so `cargo fmt` was not run.

### DIFF
--- a/plugins/codestory/scripts/codestory-mcp.cjs
+++ b/plugins/codestory/scripts/codestory-mcp.cjs
@@ -539,7 +539,7 @@ function probeLocalNavigation(resolved, projectRoot = process.cwd()) {
     );
   }
   if (parsedProbe.ok) {
-    return { ready: true };
+    return { ready: true, repaired: false };
   }
 
   const repair = runLocalNavigationReady(resolved, projectRoot, true);
@@ -572,7 +572,7 @@ function probeLocalNavigation(resolved, projectRoot = process.cwd()) {
     );
   }
   if (parsedRepair.ok) {
-    return { ready: true };
+    return { ready: true, repaired: true };
   }
   const verdict = parsedRepair.verdict || parsedProbe.verdict || null;
   const status = typeof verdict?.status === 'string' ? verdict.status : 'repair_setup';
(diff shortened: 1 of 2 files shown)


TheGreenCedar:
Coordinator review gate: keep draft.

Independent review found a blocker in the #560 boundary. The new local repair path is local-only, but after a successful local repair the launcher can still fall through to the enabled sidecar policy and schedule `ready --goal agent --repair`. That means MCP startup can still begin packet/search sidecar repair from the same path that is supposed to be local-graph-only.

Required before ready:
- Track whether startup became ready because the launcher performed the bounded local repair.
- If that repair happened, skip `scheduleSidecarRepair` for that startup path.
- Preserve existing enabled-policy behavior for an already-ready startup if that remains intentional.
- Add a regression test with sidecar policy enabled proving stale-local repair does not spawn `ready --goal agent --repair`.

A fix lane is active; no merge until that lands and review/checks pass.

Issue #578, opened 2026-06-25
Bound sidecar stage overruns in live packet/search eval


TheGreenCedar:
Parent: #569
Related: #544, #511

## Problem
The #569 diagnostic reproduced the #544 live packet/search eval miss on current `dev/codestory-next`. Exact-symbol search resolves `LiveSidecarSearch::qdrant_search` at rank 1, but broad readiness queries spend the whole sidecar budget in Zoekt and never get useful SCIP/Qdrant/graph contribution before the deadline.

This blocks #544 because the product path cannot yet prove broad-query packet/search quality with full sidecar retrieval.

## Acceptance Criteria
- Reproduce the current live eval miss or use the latest #569 transcript as the starting failing evidence.
- Enforce bounded sidecar stage behavior so one slow sidecar stage cannot consume the whole broad-query budget and starve later stages.
- Preserve deterministic fallback or diversification when an early sidecar stage overruns.
- Keep exact-symbol resolvability and broad-query packet/search quality as distinct signals.
- Report rank/stage provenance clearly enough to show whether Zoekt, SCIP, Qdrant, and graph stages contributed.
- Verification includes the focused retrieval/unit tests plus the production packet/search eval path when full sidecar retrieval is available; if live sidecar repair is unavailable, post the exact blocker and do not claim #544 is ready.
- Does not raise eval baselines, widen `k`, change fixture expectations, add new retrieval machinery, add model families/rerankers/indexes, or change product defaults without eval evidence.

## Suggested Owners
- `crates/codestory-retrieval/src/executor.rs`
- `crates/codestory-retrieval/src/planner.rs`
- `crates/codestory-retrieval/src/ranker.rs`
- `crates/codestory-runtime/src/agent/retrieval_primary.rs`
- `crates/codestory-cli/tests/packet_search_eval.rs`

## Non-Goals
- Does not merge #544 by itself.
- Does not repair Qdrant process management unless that is the direct reproduced blocker.
- Does not change package/release lanes.

TheGreenCedar:
Coordinator lane start: queued Codex worktree from origin/dev/codestory-next. Pending worktree id: local:ac6dfe0c-fc4a-4cdd-9529-9431805cc720. Expected branch after setup: codex/issue-578-sidecar-stage-overruns. Target: dev/codestory-next. Scope: enforce bounded sidecar stage behavior without changing eval baselines, k, product defaults, or adding new retrieval machinery.

[...]

Issue #569, opened 2026-06-25
Diagnose live sidecar readiness query deadline miss


TheGreenCedar:
Parent: #511
Related: #544

## Problem
Draft PR #544 is blocked by a live production packet/search eval miss that does not look like a source-level regression in the query-shape prototype.

Observed evidence from the #544 diagnostic lane:
- Live production eval failed with `recall_at_k actual=0.875 expected=1.000`.
- The missed readiness fixture expected `LiveSidecarSearch::qdrant_search`.
- Exact-symbol search returned `LiveSidecarSearch::qdrant_search` at rank 1 in `crates/codestory-retrieval/src/sidecar_search.rs:77`.
- The broad readiness query missed the symbol under live sidecar/deadline behavior, apparently because lexical-only work completed before graph/semantic contribution.
- Direct follow-up also saw `mode=unavailable, reason=zoekt_stub` in one fixture path after status/preflight had reported full.

## Acceptance Criteria
- Reproduce or explicitly retire the #544 live eval miss with current `dev/codestory-next`.
- Identify whether the miss is deadline/racing, sidecar availability, query-shape scoring, or fixture expectation drift.
- Preserve the distinction between exact-symbol resolvability and broad-query packet/search quality.
- Produce before/after live eval evidence or a no-code diagnostic comment with the exact non-code blocker.
- Do not broaden #544 or change product defaults unless the eval proves the change.

## Non-Goals
- Does not add new retrieval machinery.
- Does not merge #544 by itself.
- Does not change benchmark baselines without live production packet/search evidence.

TheGreenCedar:
Coordinator update: started a dedicated Codex diagnostic worktree for #569.

Pending worktree: `local:bb19c754-b44c-44b5-927f-cc50e2d643d1`
Expected branch: `codex/issue-569-sidecar-readiness-deadline`
Target: `dev/codestory-next`

Scope guard:
- Owns reproducing or retiring the #544 live sidecar readiness query miss.
- Does not touch active #559 cancellation, #562/#566 readiness/status, #561 hooks, or #491 release-smoke work.

TheGreenCedar:
Diagnostic update from `C:\Users\alber\.codex\worktrees\c79a\codestory` on branch `codex/issue-569-sidecar-readiness-deadline` at `f474897b` (`origin/dev/codestory-next`). No source edits.

Setup/readiness:
[...]

PR #544, opened 2026-06-24
Prototype query-shape fusion

TheGreenCedar:
## Context

Closes #511.
Refs #569, #593.

PR #593 is now merged into `dev/codestory-next`; this PR has been replayed onto `origin/dev/codestory-next` at `85bdc714384f769208af285b74ca2ae26534e10f` and is clean/mergeable on GitHub at head `c19b7fc00c25e6ab2f59aa34b7ad9f256a02c71a`.

This stays inside the existing sidecar retrieval path. It does not touch `main`, release metadata, model families, rerankers, alternate lexical indexes, product defaults, sidecar setup, Docker, cache deletion, or live packet/search evals.

## What Changed

- Added deterministic lexical provenance for the Zoekt stage via the existing `RetrievalStageKind::provenance_label` contract.
- Preserved duplicate-candidate metadata during union: file role, SCIP hop distance, and all lane provenance labels now survive when lexical/dense/graph stages identify the same file/symbol.
- Updated rank feature fusion to read existing provenance labels (`lexical_source`, `dense_anchor`, `component_report`, `graph_neighbor`, `exact`) instead of treating the retained candidate source as the only lane signal.
- Updated packet/search score breakdown reporting to expose fused rank features when the ranker populated them.
- Fixed the review blocker where non-graph rank features could export `graph > 0.0` for pure lexical/dense candidates; exported graph score is now nonzero only with graph source/provenance.
- Added focused regressions for candidate union, rank fusion, score breakdown provenance, graph-evidence export, stage metadata, and coexistence with the #593 lexical-source ranking tests.

```mermaid
flowchart LR
  A["Zoekt / Qdrant / SCIP stages"] --> B["merge_candidates"]
  B --> C["provenance-preserving candidate"]
  C --> D["rank_candidates"]
  D --> E["rank_features"]
  E --> F["packet/search score breakdown"]
```

## How To Review

- Start in `crates/codestory-retrieval/src/executor.rs` at `merge_candidates` to see duplicate candidate metadata/provenance union.
- Then read `crates/codestory-retrieval/src/ranker.rs` at `build_rank_features` and `export_rank_features` to see provenance-based fusion and graph export gating.
- Finish in `crates/codestory-runtime/src/agent/retrieval_primary.rs` at `score_breakdown_for_candidate` to confirm packet/search diagnostics report fused rank features without over-tiering pure lexical/dense candidates as graph evidence.

## Verification

[...]

Issue #511, opened 2026-06-24
Prototype query-shape ranking and deterministic fusion behind eval gates


TheGreenCedar:
Parent epic: #475

Blocked by: #510

Release lane: v0.11.20

## Purpose
Tune existing retrieval stages after the production packet/search eval baseline can measure lift by category.

## Acceptance Criteria
- [ ] Depends on #510 and does not start before that eval lane can report before/after evidence.
- [ ] Experiments with query-shape routing over existing retrieval stages.
- [ ] Experiments with deterministic candidate union and tie-break behavior.
- [ ] Reports rank provenance.
- [ ] Experiments with deterministic fusion over existing retrieval stages.
- [ ] Does not add new model families, rerankers, sparse retrieval systems, or product defaults.
- [ ] Produces before/after eval evidence by category.

TheGreenCedar:
#510 has landed via #526 on dev/codestory-next, so this lane is no longer blocked by the missing production packet/search eval baseline. Keeping this in Todo until the coordinator starts a dedicated spike/implementation lane; the dependency gate is cleared.

TheGreenCedar:
Coordinator update: starting the #511 spike/prototype lane in a dedicated Codex worktree. Pending worktree id: local:a6b4f028-e1e1-4dee-b22c-f2f491170682. Scope is bounded to existing retrieval stages behind the #510 production packet/search eval gate: query-shape routing, deterministic candidate union/tie-breaks, rank provenance, and deterministic fusion evidence by category. No new model families, rerankers, sparse retrieval systems, alternate indexes, or product defaults. No v0.11.19 or main-branch work.

TheGreenCedar:
Coordinator update: the #511 worktree is materialized and active. Thread: 019efbc1-f671-7401-bd3d-ea28a51d02a8. Worktree: C:\\Users\\alber\\.codex\\worktrees\\855c\\codestory. Branch: codex/issue-511-query-shape-fusion from current dev/codestory-next. Child preflight repaired local graph; full retrieval remains unavailable, so the lane is using local graph/direct source plus the production eval harness, not live sidecar claims.

TheGreenCedar:
Coordinator update: #511 remains parked behind the live eval proof chain, now narrowed to #600.

[...]

PR #584, opened 2026-06-25
Bound sidecar stage overruns

TheGreenCedar:
## Context

Closes #578
Refs #569, #544, #511, #585

#569 reproduced the live #544 packet/search miss: exact-symbol search resolves `LiveSidecarSearch::qdrant_search`, but broad readiness searches could spend the sidecar path in early stages and never get useful later-stage contribution.

This PR keeps the fixture contract and eval baseline unchanged. It bounds broad-query sidecar stages and records the remaining live blocker under #585.

## What Changed

- `QueryExecutor` owns an `Arc<dyn SidecarSearch>` so broad-query stages can run through a timed worker instead of blocking the whole query path.
- Broad query stage budgets are proportionally scaled when the caller supplies a smaller live budget.
- Broad-stage workers are capped by a small global in-flight guard so timed-out sidecar calls cannot accumulate without bound.
- Timed-out broad stages now record `cancel_reason: "stage_deadline"`; worker-cap misses record `stage_worker_limit`; later stages still run deterministically when a stage overruns.
- Exact/path queries keep the direct execution path and the previous between-stage sequence deadline behavior, so they stop starting later stages after `total_budget_ms` is spent.
- Strict batch query execution shares the same sidecar handle shape.

```mermaid
flowchart LR
  A["Broad query"] --> B["Acquire worker slot"]
  B -->|"slot available"| C["Timed stage worker"]
  B -->|"cap reached"| D["stage_worker_limit trace"]
  C -->|"stage overruns"| E["stage_deadline trace"]
  D --> F["Later stages still considered"]
  E --> F
  F --> G["Qdrant/SCIP candidates can contribute"]
```

## How To Review

- Start in `crates/codestory-retrieval/src/executor.rs`.
- Check that `run_stage_bounded` only applies timed workers to broad query shapes.
- Check `BroadStageWorkerPermit` for the bounded in-flight guard.
- Check `non_broad_queries_stop_between_stages_after_total_deadline` for restored exact/path deadline behavior.
- Check `broad_query_stage_deadline_preserves_later_sidecar_contribution` and `broad_query_slow_scip_expand_still_allows_qdrant_contribution` for the #578 overrun cases.
- Check `crates/codestory-retrieval/src/query.rs` for the minimal `Arc<dyn SidecarSearch>` follow-through in single and strict batch query paths.

## Verification

Review follow-up at `4ae066fe90220ad916a4a5755e8f04daeaf76194`:

- `cargo fmt`
  - passed
- `cargo test -p codestory-retrieval executor`
  - passed: 17 passed
- `cargo test -p codestory-retrieval strict_batch`
  - passed: 2 passed
- `git diff --check`
  - passed

Earlier #584 verification before the review follow-up:

- `powershell -ExecutionPolicy Bypass -File scripts\codex-worktree-setup.ps1`
  - local graph ready: `files=275`, `nodes=116865`, `edges=98039`, `errors=0`
  - sidecar retrieval ready after index refresh: `retrieval_mode=full`
- `cargo test -p codestory-cli --test packet_search_eval`
  - passed: 7 passed, 1 ignored
- Exact-symbol control:
  - `target\debug\codestory-cli.exe search --query "LiveSidecarSearch::qdrant_search" --limit 10 --refresh none --why --format json`
  - `LiveSidecarSearch::qdrant_search` resolved at rank 1
- Broad readiness diagnostic after stage bounding:
  - full retrieval, `cancel_reason=stage_deadline`
  - `stage1_zoekt_lexical`: `deadline_ms=120`, `elapsed_ms=120`, `cancel_reason=stage_deadline`, `candidates_added=0`
  - `stage2_scip_expand`: `deadline_ms=180`, `elapsed_ms=180`, `cancel_reason=stage_deadline`, `candidates_added=0`
  - `stage1b_qdrant_semantic`: `elapsed_ms=37`, `candidates_added=40`
- Live production eval remained failing before this review follow-up:
  - `cargo test -p codestory-cli --test packet_search_eval -- --ignored --nocapture packet_search_eval_live_runs_production_cli_path`
  - failed after 184.27s: `recall_at_k regressed: actual=0.500 expected=1.000 tolerance=0.000`

## Risk

Draft because #544 is not ready. This removes the broad-query starvation mode and fixes the review blockers around detached worker accumulation and non-broad sequence deadlines, but the live eval miss remains parked on #585.

This PR does not raise baselines, widen `k`, change fixture expectations, add retrieval machinery, add model families/rerankers/indexes, or merge #544.

### DIFF
--- a/crates/codestory-retrieval/src/executor.rs
+++ b/crates/codestory-retrieval/src/executor.rs
@@ -12,8 +12,8 @@ use codestory_store::RetrievalIndexManifest;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::path::Path;
-use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, mpsc};
 use std::time::{Duration, Instant};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -74,7 +74,7 @@ pub struct QueryResult {
 /// The executor is fail-closed: live degraded modes return an error instead of serving partial
 /// results. Tests may use `mode_override`, but product callers should rely on live health probes.
 pub struct QueryExecutor<'a> {
-    pub sidecars: &'a dyn SidecarSearch,
+    pub sidecars: Arc<dyn SidecarSearch>,
     pub cache: &'a mut RetrievalCache,
     pub manifest: Option<RetrievalIndexManifest>,
     pub file_roles: HashMap<String, codestory_store::FileRole>,
@@ -122,12 +122,15 @@ impl<'a> QueryExecutor<'a> {
         }
 
         let mut plan = crate::planner::plan_query(&features, mode);
+        let planned_budget_ms = plan.stages.iter().map(|stage| stage.budget_ms).sum::<u64>();
         if let Some(budget) = total_budget_ms {
+            if is_broad_query(features.shape) && budget < planned_budget_ms {
+                scale_stage_budgets(&mut plan.stages, budget);
+            }
             plan.total_budget_ms = budget;
         }
 
         let started = Instant::now();
-        let deadline = started + Duration::from_millis(plan.total_budget_ms);
         let mut candidates = Vec::new();
         let mut stage_traces = Vec::new();
 
@@ -136,7 +139,6 @@ impl<'a> QueryExecutor<'a> {
             &plan.stages,
             &mut candidates,
             &mut stage_traces,
-            deadline,
             StageSequenceOptions {
                 stop_marginal_gain_threshold: Some(plan.stop_marginal_gain_threshold),
                 stop_after_low_gain_streak: plan.stop_after_low_gain_streak,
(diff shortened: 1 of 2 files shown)


TheGreenCedar:
Coordinator review state: #584 is useful but stays draft until checks and review finish. It appears to address #578's sidecar-stage starvation slice, but it does not unblock #544 by itself: the PR's own live production eval still fails (
ecall_at_k=0.500 vs 1.000). I am splitting the remaining ranking/window blocker under #569 so this PR can be judged narrowly.

TheGreenCedar:
Coordinator split: created #585 for the remaining ranking/window quality blocker after this PR's Qdrant contribution evidence. That keeps #584 review scoped to whether the sidecar-stage overrun bound is correct and safe.

TheGreenCedar:
Coordinator blocker from independent review. Keep #584 draft.

Findings to fix before ready/merge:

1. Timed-out stage workers are detached and can accumulate. `executor.rs` leaves timed-out sidecar calls running, and strict batch can multiply this. Add a small in-flight cap/guard for spawned broad-stage workers, or make sidecar calls observe a cooperative deadline/cancel token.

2. Exact/path queries kept the direct stage path, but the old between-stage sequence deadline check was removed. Preserve the previous deadline behavior for non-broad shapes so exact/path queries still stop starting later stages after `total_budget_ms` is spent.

3. The focused regression only covers slow Zoekt. Add the same focused coverage for slow `Stage2ScipExpand` with Qdrant still contributing, since #578 acceptance is about one slow sidecar stage not starving later stages.

After the fix, rerun focused retrieval tests and keep the live-eval non-claim intact: #544 remains parked on #585 unless production eval passes.

TheGreenCedar:
Review follow-up pushed: `4ae066fe90220ad916a4a5755e8f04daeaf76194`.

Fixed the three blockers:

- Added a small global in-flight guard for spawned broad-stage workers. If the cap is hit, the stage records `stage_worker_limit` instead of spawning more detached work.
- Restored the between-stage total deadline check for non-broad shapes while keeping exact/path stage calls direct.
- Added focused slow `Stage2ScipExpand` coverage proving Qdrant still contributes after graph expansion times out.

Verification on this follow-up:

- `cargo fmt` passed
- `cargo test -p codestory-retrieval executor` passed: 17 passed
- `cargo test -p codestory-retrieval strict_batch` passed: 2 passed
- `git diff --check` passed

Live-eval non-claim remains: #544 is still parked on #585; this follow-up does not merge or unblock #544.

Issue #611, opened 2026-06-25
Add operator-approved sidecar GC apply path


TheGreenCedar:
Parent: #496
Related: #569, #582, #587, #593, #544

## Problem
The read-only sidecar inventory now reports owned stale/orphaned safe candidates, but the release-adjacent packet/search proof is still blocked by Docker address-pool exhaustion. Operators need an explicit, guarded apply path for CodeStory-owned safe candidates instead of manual Docker/cache cleanup.

The coordinator proof after #609 saw:
- `ready --goal agent --repair` failed with `reason=docker_address_pool_exhausted`.
- Inventory was read-only only: 98 namespaces, 28 live, 26 incomplete, 44 orphaned/safe candidates.
- No cleanup/down/prune/delete/restart command was run.

## Acceptance Criteria
- Adds an explicit sidecar GC apply command or mode that consumes the existing inventory safe-candidate classification.
- Requires an unmistakable opt-in flag such as `--apply`; dry-run remains the default.
- Deletes only CodeStory-owned safe candidates from inventory; live namespaces and unknown ownership remain blocked.
- Prints exactly what was removed and what stayed blocked, in JSON and human-readable output.
- Includes a per-namespace refusal reason when cleanup is blocked.
- Does not run Docker prune, broad Docker cleanup, or host/global network deletion.
- Adds tests for dry-run no-deletion, apply removes only safe candidates, live resources are refused, unknown ownership is refused, and partial failures remain visible.
- Updates the sidecar ops docs with the safe operator sequence.

## Non-Goals
- Does not auto-run cleanup from readiness, hooks, MCP startup, or packet/search evals.
- Does not change retrieval ranking or query fusion.
- Does not require CodeStory on PATH.
- Does not bypass owner/operator approval.

TheGreenCedar:
Coordinator update: the queued implementation worktree has surfaced and is active.

Thread: `019efe42-1297-74c3-8378-9dbbb6cc4ffb`
Worktree: `C:\Users\alber\.codex\worktrees\b29d\codestory`
Branch: `codex/issue-611-sidecar-gc-apply`

Current observed progress from the child thread: code/docs/tests are being patched, focused retrieval inventory tests have passed, and the worker is adding a parser test before opening the draft PR. Coordinator has not run cleanup or mutated Docker/cache state.

TheGreenCedar:
[...]

PR #613, opened 2026-06-25
[codex] add sidecar gc apply

TheGreenCedar:
## Context

Closes #611.

The read-only sidecar inventory already classifies stale/orphaned CodeStory-owned namespaces as safe candidates. Operators needed an explicit apply path for those candidates without readiness-triggered cleanup, Docker prune, or broad host cleanup.

## What Changed

- Added `retrieval inventory --apply` and the `sidecar inventory --apply` alias.
- Kept dry-run inventory as the default when `--apply` is omitted.
- Added GC output with `removed` and `blocked` namespace lists for JSON and Markdown.
- Applied cleanup only to inventory safe candidates, with live, incomplete, unknown-owner, and partial-failure namespaces reported as blocked.
- Updated the sidecar ops runbook with the dry-run first, operator-approved apply sequence.

## How To Review

- Start in `crates/codestory-retrieval/src/inventory.rs` and confirm apply uses the existing inventory classification.
- Check `crates/codestory-cli/src/args.rs` and `crates/codestory-cli/src/retrieval.rs` for the opt-in CLI surface and output shape.
- Check `docs/ops/retrieval-sidecars.md` for the operator sequence and non-goals.

## Verification

- `cargo test -p codestory-retrieval inventory::tests -- --nocapture`
- `cargo test -p codestory-cli retrieval_inventory_apply_is_explicit_opt_in -- --nocapture`
- `cargo test -p codestory-cli inventory -- --nocapture`
- `cargo fmt --check`
- `git diff --check`

## Risk

The apply path can remove local CodeStory-owned cache artifacts and explicitly matched sidecar resources when an operator passes `--apply`. It does not run Docker prune, broad Docker cleanup, host/global network deletion, or readiness-triggered cleanup. I did not run a live `--apply` CLI smoke against host Docker; behavior is covered with controlled unit inventory inputs to avoid deleting workstation resources during verification.

### DIFF
--- a/crates/codestory-cli/src/args.rs
+++ b/crates/codestory-cli/src/args.rs
@@ -828,6 +828,11 @@ pub(crate) struct RetrievalStatusCommand {
 pub(crate) struct RetrievalInventoryCommand {
     #[command(flatten)]
     pub(crate) project: ProjectArgs,
+    #[arg(
+        long,
+        help = "Apply cleanup for CodeStory-owned inventory safe candidates. Omit for dry-run inventory only."
+    )]
+    pub(crate) apply: bool,
     #[arg(long, value_name = "FORMAT", value_parser = parse_read_output_format, default_value = "json")]
     pub(crate) format: OutputFormat,
     #[arg(long, value_name = "PATH")]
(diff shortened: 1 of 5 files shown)


TheGreenCedar:
[P1] Block merge until Docker resource ownership is revalidated for apply.

`match_resource` currently accepts a Docker resource when `dev.codestory.namespace` matches the candidate namespace, without checking `dev.codestory.owner`. The later unknown-owner guard only checks resource ownership when there is no state file. With an owned stale/orphaned state file, a stopped foreign-owned resource carrying the same namespace label could be included in a safe candidate and removed by `--apply`.

Please require CodeStory ownership for matched resources before apply, except for the intentional unlabeled compose-network case tied to an owned state/compose project, and add focused regression coverage for owned state plus foreign-owner same-namespace resource.

Issue #620, opened 2026-06-25
Use stable shared agent sidecar run id for local repair


TheGreenCedar:
Parent: #619
Related: #569, #611, #613, #617

## Problem
Normal local agent sidecar repair can mint a fresh agent run namespace on each attempt when no run id is provided. That can accumulate CodeStory-owned Docker compose projects/networks and eventually exhaust Docker address pools.

The first product fix should reduce sidecar creation at the source. Cleanup/inventory remains an operator escape hatch, not the ordinary recovery path.

## Acceptance Criteria
- Normal local `ready --goal agent --repair` uses one stable project-scoped agent run id by default when the operator did not explicitly provide a run id.
- Plugin/status `sidecar_setup.next_repair_command` points at the stable shared agent run id for local agent repair.
- Explicit `--run-id` remains honored for eval, CI, and isolated proof lanes.
- Before starting compose for the shared namespace, status/repair output makes the target namespace/run id visible enough for operators to understand reuse.
- Existing full-retrieval proof identity remains strict: no full packet/search claim without a current finalized manifest for the target project/code identity.
- Inventory/cleanup remains explicit and operator-approved; no automatic Docker prune, deletion, restart, or `inventory --apply`.
- Include focused tests for default shared run id, explicit run id override, status/repair command text, and no change to eval fixed-run-id behavior.

## Non-Goals
- Does not implement the full shared sidecar service architecture.
- Does not change packet/search ranking or eval baselines.
- Does not auto-clean Docker resources.
- Does not merge #593/#544 or claim live proof.

TheGreenCedar:
Coordinator lane queued.

Pending Codex worktree: `local:591d3b76-4a15-4a7a-8949-025ecc49a62a`

Scope: implement the smallest #620 slice on `dev/codestory-next`: stable shared agent run id for normal local repair, status/repair command text, explicit run-id override preserved, focused tests, draft PR closing #620. No cleanup, Docker prune/delete/restart, cache deletion, or `inventory --apply`.

TheGreenCedar:
Coordinator update: #620 worktree surfaced.

Thread: `019efeaf-beb7-7e00-a302-f0eaca1b72bb`
Worktree: `C:\Users\alber\.codex\worktrees\dd94\codestory`

Implementation lane is active on the stable shared agent sidecar run id slice. No cleanup or host sidecar mutation was delegated.

TheGreenCedar:
[...]

PR #623, opened 2026-06-25
Use stable shared agent run id for repair

TheGreenCedar:
## Context

Closes #620.

Normal local agent repair could create a fresh agent sidecar namespace when no run id was provided. That made repeated `ready --goal agent --repair` attempts more likely to accumulate CodeStory-owned compose projects/networks. The first fix is to reuse one project-scoped agent namespace by default while preserving explicit run ids for isolated proof lanes.

## What Changed

- Added `shared-agent` as the default agent sidecar run id in `codestory-retrieval`.
- Added `ready --run-id <ID>` and threaded it through agent readiness check/repair paths.
- Made agent readiness lanes and repair hints point at the shared agent run id by default instead of the old missing-run sentinel.
- Updated plugin `sidecar_setup.next_repair_command`, the plugin launcher env command, and enabled-policy `recommended_next_calls` to include the shared run id.
- Kept the live packet/search eval isolated by using `packet-search-eval` for readiness/status args and by setting `CODESTORY_RETRIEVAL_PROFILE=agent` plus `CODESTORY_SIDECAR_RUN_ID=packet-search-eval` on every live eval subprocess, including `search` and `packet`.
- Kept strict full-retrieval identity unchanged: readiness still depends on strict sidecar status/finalized manifest checks before packet/search become full.

```mermaid
flowchart LR
    Ready[ready --goal agent] --> Runtime[agent runtime]
    Runtime --> Default[run_id shared-agent]
    Runtime --> Explicit[--run-id override]
    Default --> Status[status/repair hints]
    Explicit --> Status
    Status --> Strict[strict sidecar status]
    Strict --> Full[full packet/search only with current manifest]
```

## How To Review

- Start in `crates/codestory-retrieval/src/config.rs` for the default run-id behavior.
- Then review `crates/codestory-cli/src/main.rs` and `crates/codestory-cli/src/readiness.rs` for ready/status command plumbing.
- Check `crates/codestory-cli/src/stdio_transport.rs` and `plugins/codestory/scripts/codestory-mcp.cjs` for plugin status/launcher repair command text.
- Check `crates/codestory-cli/tests/packet_search_eval.rs` for the fixed eval/proof run id and per-subprocess env identity.
- Tests in `ready_command.rs`, `stdio_protocol_contracts.rs`, `plugin-static.test.mjs`, `packet_search_eval.rs`, and `cli_golden_path.rs` show the expected operator-facing command text.

## Verification

- `cargo fmt --check`
- `git diff --check`
- `cargo test -p codestory-retrieval agent_profile_default_runtime_reuses_project_shared_run -- --nocapture`
- `cargo test -p codestory-retrieval agent_profile_run_id_reuses_namespace_state_and_ports -- --nocapture`
- `cargo test -p codestory-cli --test ready_command ready_command_emits_compact_verdicts_and_filters_goal -- --nocapture`
- `cargo test -p codestory-cli --test stdio_protocol_contracts resources_read_status_recommends_sidecar_repair_when_policy_enabled -- --nocapture`
- `cargo test -p codestory-cli --test cli_golden_path doctor_next_commands_stop_at_retrieval_repair_when_sidecar_is_not_full -- --nocapture`
- `cargo test -p codestory-cli --test cli_golden_path agent_preflight_reports_local_graph_when_retrieval_is_degraded -- --nocapture`
- `cargo test -p codestory-cli agent_readiness_runtime_does_not_collapse_to_local_without_agent_run -- --nocapture`
- `cargo test -p codestory-cli agent_goal_readiness_keeps_handoff_mismatch_when_agent_lane_is_not_full -- --nocapture`
- `cargo test -p codestory-cli --test retrieval_bootstrap_contracts agent_profile_bootstrap_status_and_down_are_project_isolated -- --nocapture`
- `cargo test -p codestory-cli --test packet_search_eval packet_search_live_eval_uses_fixed_run_id -- --nocapture`
- `node --test --test-name-pattern "mcp launcher prefers a checksummed managed cli without PATH" plugins\codestory\tests\plugin-static.test.mjs`

No live `ready --goal agent --repair`, Docker prune/delete/restart, cache deletion, or `inventory --apply` was run. The bootstrap contract above uses temp test state and `--skip-compose` to verify explicit run-id isolation.

## Risk

- Existing agent sidecars with older ad hoc run ids remain valid when explicitly selected, but normal repair/status now points at `shared-agent`.
- Operators relying on implicit latest-agent autodiscovery for local repair will now need to pass `--run-id` explicitly for that isolated namespace.
- The live eval uses `packet-search-eval` for readiness/status isolation and sets the matching agent profile/run-id env for search and packet subprocesses; the ignored live packet/search test itself was not run in this patch pass.


TheGreenCedar:
Coordinator review found blockers after draft PR #623 opened.

Current state:
- #623 is draft and remains the right lane for #620.
- Issue-link guard passed because the PR body contains `Closes #620`, though GitHub's computed closing reference is empty on the dev-targeted PR.
- Independent review found two fix-before-ready items: plugin-managed status repair command still omits `--run-id`, and the live packet/search eval path now needs an explicit proof/eval run id instead of inheriting the shared default.

I routed those findings back to the implementation thread. Keep #623 draft until the branch is patched, focused tests are updated, and checks settle.

TheGreenCedar:
Coordinator follow-up: #623 stays draft.

The first blocker patch fixed plugin repair text and readiness/status command contracts, but a remaining eval identity gap surfaced during re-check: the live eval uses `--run-id packet-search-eval` for readiness/status, while the later `search` and `packet` subprocesses can still use the ambient auto sidecar because each CLI call is a fresh process.

I routed the smallest follow-up back to the implementation thread: make the live eval harness pass the fixed agent run id/profile to all subprocesses in that eval path, and extend the non-ignored command-contract test to prove it.

Issue #955, opened 2026-07-12
Complete the atomic publication crash and cancellation matrix


TheGreenCedar:
Parent: #910

## Scope

Complete the fail-closed crash and cancellation evidence for atomic full and
incremental publication.

## Acceptance criteria

- Enumerate every publication transition for full and incremental refresh.
- Add deterministic fault injection for publication identity persistence,
  search generation build/validation/completion, catalog-lock acquisition,
  staged database replacement, and runtime in-memory publication in addition
  to the existing seven incremental SQLite boundaries.
- Add cancellation coverage at every phase and real process crash/kill coverage
  at the filesystem publication boundary where deterministic cross-process
  proof is required.
- Prove readers observe only the previous complete generation or the fully new
  complete generation after every injected failure.
- Prove restart recovery preserves the last complete publication and safely
  cleans or retains incomplete staging artifacts.
- Keep the matrix table-driven and production code free of holdout literals.
- Update the changelog and the nearest durable testing documentation.

## Verification

- Focused runtime/store matrix tests
- Real two-process crash/restart tests on supported CI operating systems
- Full refresh and incremental fidelity/regression suites
- Focused clippy with warnings denied

Refs #899

TheGreenCedar:
# Review #955

- Post-commit cancellation currently returns rollback-shaped errors after the live DB changed.
- Runtime search state can publish before the incremental marker clears.
- Crash proof runs after replacement cleanup and accepts any child failure, so it can false-pass.
- Cancellation injection bypasses production observation; marker clear is missing from the transition table.
- Rebase onto current `dev/codestory-next` before publication.

Source-backed review of the uncommitted worktree; no Cargo commands run.

TheGreenCedar:
# Re-review #955 / PR #973

- Cancellation still returns `cancelled` after the live database has committed.
- Runtime search state publishes and clears `is_indexing` before the incremental marker clears.
- Crash testing occurs after restore/cleanup and can false-pass through the child's fallback panic.
- Cancellation hooks return errors themselves instead of proving production observes cancellation.
- Marker-clear cancellation remains absent from the transition matrix.
- Rebase onto the current `dev/codestory-next` before publication.

Exact-head source review of `db8b3ff7`; no Cargo commands run.

Issue #958, opened 2026-07-12
Activate the release-evidence gate with a live v0.15 baseline and candidate


TheGreenCedar:
Parent: #922

## Scope

Activate the attested release-evidence gate with a live v0.15 baseline and
candidate only after the infrastructure merged in #953 is proven on the real
release environment.

## Acceptance criteria

- Provision and document the labeled runner, drill manifest, exact embedding
  model, full retrieval sidecars, and required machine capacity.
- Produce and review a live baseline for a release-eligible machine profile,
  including a full clean source SHA, artifact hashes, runtime identities, all
  seven budgets, owner/date/rationale, and measured storage bytes.
- Run the manual evidence workflow for a v0.15 release-candidate SHA and retain
  provisioning, raw repo, raw packet, candidate, and decision artifacts.
- Resolve rejected metrics or use an exact, scoped, expiring approval followed
  by prior-run re-evaluation; no self-ratcheting reference update is allowed.
- Wire the proven reusable workflow into both release entry points only after a
  live approved profile and passing release candidate exist, without creating
  a rollout deadlock.
- Audit runtime, non-Rust, configuration, generated, and external corpus
  indirection; resolve or separately track any guard gap rather than overstating
  the Rust production-corpus boundary.
- Update operator/testing documentation and the changelog with the activated
  release contract and recovery procedure.

## Verification

- Live source-backed baseline and candidate artifacts from the provisioned
  release environment
- Passing re-evaluation of the exact candidate artifacts
- Release and auto-release policy checks with the reusable gate activated
- Negative tests for missing/stale/mismatched/expired evidence
- Documentation links and `git diff --check`

Close #922 only after this live proof and release-path activation are verified.

Refs #899

TheGreenCedar:
Draft activation wiring is published as #1078. Ordinary PR package proof remains unsigned; Developer ID signing/notarization is confined to the main release workflow.

Protected run 29301774689 is retained as rejected evidence. It proved full-sidecar lifecycle, nine successful packet invocations, and clean proof-owned teardown, but it is not release-eligible: Axios missed required symbols/claims, Redis retained 14 unresolved candidates, and Ripgrep returned partial evidence with missing entrypoint/search files. No baseline, candidate, or release decision was promoted.

#1078 remains draft until those product-quality failures are resolved and a distinct passing baseline/candidate pair exists. Merging the activation wiring before that would intentionally deadlock releases.

TheGreenCedar:
[...]

Issue #956, opened 2026-07-12
Record post-publication repo-scale timing evidence


TheGreenCedar:
Parent: #910

## Scope

Produce the missing post-atomic-publication repo-scale timing evidence without
weakening or fabricating the established benchmark contract.

## Acceptance criteria

- Run the ignored `codestory_repo_e2e_stats` harness against the current
  `dev/codestory-next` implementation using the required real-repo drill cases
  and full sidecar profile.
- Append the emitted stats row to
  `docs/testing/codestory-e2e-stats-log.md` with the exact source commit,
  machine/profile provenance, graph and semantic timings, and any declared
  repeat-refresh baseline comparison.
- Preserve raw output as review evidence and explain any unavailable metric or
  environmental deviation explicitly.
- Do not hand-author benchmark numbers or lower thresholds to obtain a pass.
- Run the release-evidence evaluator when the #922 infrastructure is available,
  while keeping this child independently truthful if activation is still
  pending.

## Verification

- Release CLI build
- Ignored repo-scale stats harness with `--nocapture`
- Documentation link check and `git diff --check`

Refs #899

TheGreenCedar:
## Published evidence-retention checkpoint

The reviewed #956 preparation is available on branch `codex/956-live-drill-evidence` at `226b32f`.

- Real-repo drills can opt into a workflow-owned persistent output directory; local runs keep the temporary default.
- The release-candidate workflow selects the dedicated Linux ARM64 runner, enables explicit CPU embeddings for both producers, uses locked dependencies, tees raw output without masking failures, and uploads the durable drill tree.
- Drill, full-sidecar, and release-promotion claims remain separate. No baseline or release gate is activated by this branch.
- Focused Rust tests, workflow policy, `actionlint`, docs links, formatting, and diff checks pass.
- Independent review found no P1/P2 issues.

Do not spend the live run yet. First merge and re-review the isolated #1064 runner, then freeze the final runtime-affecting dev head. The exact clean head must retain stats JSON, raw Cargo output, the complete drill tree, and provisioning/hash evidence.

TheGreenCedar:
Implementation checkpoint:

- Worktree: `issue-956-live-drill-evidence-v2`
- Branch: `codex/956-live-drill-evidence-v2`
- Base: `4aece7bc65547abf993aa33581ffb30e46b97898`
- Head: `b387990f563183b550b9eaed24fccab1a4c94766`
- Role: implementation
[...]

Issue #910, opened 2026-07-10
Make incremental indexing and snapshot publication atomic


TheGreenCedar:
Parent: #899

## Finding

The live incremental database, edge resolution, grounding snapshots, and search snapshots do not share one commit boundary. A cancellation or late error can publish a graph state that no completed run owns.

## Evidence

- `crates/codestory-runtime/src/lib.rs:10718`
- `crates/codestory-indexer/src/lib.rs:1084`
- `crates/codestory-store/src/storage_impl/mod.rs:1192`
- Full refresh already stages and promotes a separate database

## Acceptance Criteria

- [x] Incremental work writes to an isolated staging/savepoint boundary.
- [x] Projection, cleanup, resolution, semantic docs, and derived snapshots finalize before publication.
- [x] Cancellation/error discards staging and leaves last-known-good state readable.
- [x] Publication has one durable generation/run identity.
- [x] Fault injection covers every transition.
- [x] The v0.14.3 incomplete-run compatibility marker migrates safely.

## Verification

- Crash/cancel matrix across every phase
- Concurrent readers see only old or fully new generation
- Full and incremental fidelity/regression suites
- Repo-scale timing evidence

## Non-Goals

- Do not retain partial-success semantics for production publication.
- Do not rewrite the graph contracts unnecessarily.

## Metadata

- Release: v0.15.0
- Area: Index architecture
- Priority: P1
- Wave: Next

## Implementation Slices

- [x] #936 - staged incremental core database and grounding snapshots
- [x] #941 - durable publication generation and run identity
- [x] #943 - generation-scoped persisted search publication
- [x] #945 - staged semantic document completion
- [x] #946 - serve the last complete publication during concurrent refresh
- [x] #955 - complete the remaining publication crash/cancel transition matrix
- [x] #956 - record post-implementation repo-scale timing evidence

## Workstreams

- [x] #936 — stage incremental graph and grounding snapshots
- [x] #941 — persist publication generation identity
- [x] #943 — publish persisted search by core generation
- [x] #945 — finalize semantic documents before core publication
- [x] #955 — complete crash and cancellation matrix
- [x] #975 — reopen storage after cross-process refresh handoff
- [x] #956 — record post-publication repo-scale timing evidence

TheGreenCedar:
Closeout audit on current `dev/codestory-next` confirms #940, #942, #944,
#947, and #948 are merged and their commits are ancestors of the branch. Atomic
[...]

PR #973, opened 2026-07-12
test publication fault matrix

TheGreenCedar:
## Context

#910 established staged core/search publication, but transition coverage did
not deterministically exercise every identity, search, lock, database-swap, and
runtime-cache failure/cancellation boundary or prove recovery after an actual
process abort.

Closes #955
Refs #910
Refs #956
Refs #975
Refs #899

## What Changed

- Added a test-only fault router and table-driven full/incremental matrix across
  identity, search build/validation/completion, catalog locking, marker
  completion, database replacement, and runtime cache publication.
- Added cancellation checkpoints before the durable point of no return. A
  cancellation or injected cache fault after commit completes the prepared
  generation instead of returning a retryable ordinary failure.
- Clear incomplete-run markers in the staged database before publication, so
  readers never observe an absent generation between old and complete-new.
- Recover a genuinely interrupted SQLite restore from its backup. A persistent
  promotion lock distinguishes abandoned recovery state from a healthy writer,
  preventing concurrent readers from restoring or deleting an active backup.
- Added a child-process abort after live mutation and before backup cleanup,
  restart recovery, active-promoter reader-open, cleanup, and retry proof.
- Added the focused qualification jobs to cross-platform Rust CI and documented
  the contract.

## How To Review

Start with `promote_staged_snapshot` and `recover_interrupted_promotion`: the
promotion lock must cover recovery, backup creation, live restore, and cleanup.
Then review marker completion before staged publication and the post-commit
runtime-cache semantics. Finally inspect the test-only fault router and matrix.

## Verification

- `cargo test -p codestory-runtime incremental_publication_transitions_fail_or_cancel_atomically -- --nocapture` — passed (16 transition cases)
- `cargo test -p codestory-runtime full_publication_transitions_fail_or_cancel_atomically -- --nocapture` — passed (14 applicable transition cases)
- `cargo test -p codestory-runtime full_recovery_marker_completion_fault_preserves_fenced_live_generation -- --nocapture` — passed
- `cargo test -p codestory-store staged_promotion_abort_recovers_old_or_complete_new_and_cleans_artifacts -- --nocapture` — passed
- `cargo test -p codestory-store reader_open_during_healthy_promotion_does_not_recover_active_backup -- --nocapture` — passed
- `cargo clippy -p codestory-store -p codestory-runtime --all-targets -- -D warnings` — passed
- `cargo fmt --all` and `git diff --check` — passed

The ignored repo-scale timing/evidence lane still requires the live model and
drill manifest unavailable in this environment. No evidence row was fabricated;
#956 and #958 retain that release-proof scope.

## Risk

The fault injection is test-only. The production change adds one persistent
lock file beside the live database and uses the workspace's existing `fs4`
dependency. Readers may briefly wait on SQLite while a healthy restore is in
flight, but they cannot consume or roll back an active writer's backup. No UI
surface changed.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -122,3 +122,9 @@ jobs:
 
       - name: Prove multi-project embedding endpoint isolation
         run: cargo test --locked -p codestory-cli --test stdio_protocol_contracts multi_project_stdio_keeps_project_embedding_endpoints_isolated_across_a_b_a -- --nocapture
+
+      - name: Prove publication transition failure and cancellation semantics
+        run: cargo test --locked -p codestory-runtime publication_transitions_fail_or_cancel_atomically -- --nocapture
+
+      - name: Prove staged database recovery after process abort
+        run: cargo test --locked -p codestory-store staged_promotion_abort_recovers_old_or_complete_new_and_cleans_artifacts -- --nocapture
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,15 @@
   client, and a cross-platform A/B/A stdio contract proves two trusted projects
   cannot cross-route embedding markers or vectors.
 
+- Added a table-driven full and incremental publication fault matrix covering
+  identity persistence, search build/validation/completion, catalog locking,
+  staged database replacement, incremental marker completion, and runtime cache
+  publication. Every boundary now has deterministic failure and cancellation
+  proof, including successful completion when cancellation arrives after the
+  database replacement point of no return. A store-level child process abort
+  after live mutation and before backup cleanup proves retry recovery and
+  artifact cleanup. Promotion-scoped locking prevents healthy backups from
+  being mistaken for abandoned crash recovery state by concurrent readers.
 - Added parser-backed Express route extraction for static module-scope
   registrations on source-ordered app and router bindings constructed from
   explicit `express` imports or `require("express")`. JavaScript, TypeScript,
@@ -104,6 +113,9 @@
   when its complete publication stayed stable, and keyed the cache by durable
   publication identity so concurrent publication cannot produce or retain a
   mixed-generation response.
+- Completed the durable incremental marker before exposing prepared runtime
+  search state or reporting indexing idle. Marker failure now leaves runtime
+  state cleared and the new database generation fenced from readers.
 - Corrected the Codex managed-platform guide to list the restored macOS x64 CLI
   while retaining its no-managed-accelerated-sidecar boundary.
 - Retried packaged-proof temporary directory removal after transient executable
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -573,6 +573,7 @@ version = "0.14.3"
 dependencies = [
  "anyhow",
  "codestory-contracts",
+ "fs4",
  "parking_lot",
  "rusqlite",
  "serde",
--- a/crates/codestory-cli/tests/architecture_contracts.rs
+++ b/crates/codestory-cli/tests/architecture_contracts.rs
@@ -417,7 +417,7 @@ fn runtime_snapshot_lifecycle_flows_through_store_snapshot_surface() {
 }
 
 #[test]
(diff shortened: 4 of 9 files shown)


TheGreenCedar:
Independent review of exact head `2238cee6` remains blocking.

- **P1 — incomplete incremental publication is accepted.** The fault/cancel
  matrix accepts a live raw generation whose incomplete marker remains and for
  which `get_complete_index_publication()` returns `None`
  (`crates/codestory-runtime/src/lib.rs:19806-19835`). Production can return
  cancellation after the database swap (`:8868-8889`). This contradicts the
  required old-or-complete-new invariant.
- **P1 — search state becomes visible before publication completion.** Prepared
  runtime search is published and `is_indexing` is cleared at `:8892-8895` /
  `:12467-12477`, before the incomplete marker clears at `:8950`, leaving a
  serving window over a fenced database.
- **P1 — cancellation proof bypasses production observation.** The test hook
  cancels and immediately returns its own error (`:4213-4218`), so the matrix
  does not prove the production transition observes cancellation safely.
- **P1 — marker-clear cancellation is absent.** The transition catalog at
  `:19719-19727` omits marker clearing, and `finish_incremental_run_marker` has
  no cancellation boundary.
- **P1 — crash proof runs after recovery is already complete and can
  false-pass.** The crash hook runs only after `staged.publish()` returns
  (`:11660-11673`, `:12098-12111`), after Store promotion/backup cleanup
  (`crates/codestory-store/src/storage_impl/mod.rs:2145-2186`). The child
  accepts any non-success and a fallback panic (`runtime/lib.rs:20002-20003`),
  so it does not prove crash recovery during the vulnerable transition.

The current Linux failure is also real: the pre-lock inspection runtime is
reused after lock acquisition (`crates/codestory-cli/src/main.rs:2432-2479`),
allowing generation `+2` and returning `cache_busy`. That root cause belongs to
#975; its minimal reopen-after-lock fix must land before this matrix can be
green.

Do not merge this head. Rework the production cancellation/publication order
and crash injection points, then rerun independent review against the new
head.

TheGreenCedar:
Correction to the #975 dependency diagnosis after PR #981's Linux full-suite
reproduction:

The `cache_busy` is not caused by a stale pre-lock runtime or a second
generation. Summary reads already open fresh storage and lock acquisition does
not wait. One incremental refresh exposes two live complete-publication edges:

1. old complete publication → `None` when the staged database is published
   with its incomplete marker still set;
2. `None` → new complete publication when the marker is cleared afterward.

This confirms the publication-order P1 already reported here. Fix #973 by
finishing the marker in staged storage after validation/search completion and
before database replacement, then remove the live post-publication marker
clear. Move the MarkerCompletion fault/cancellation boundary before
DatabaseReplacement and add a deterministic barrier proving live remains old
until one atomic new complete generation appears.

#975 and ineffective PR #981 are closed as a misdiagnosed child. The other four
P1 findings on this PR remain unchanged.

TheGreenCedar:
Read-only review of the current uncommitted repair snapshot (HEAD `2238cee6`,
six modified files, diff hash `3265d8a013030297173324a817078c5929a3d45d`):

Resolved in the draft:

- prepared runtime search is published only after marker completion;
- cancellation injection now lets production observe the cancelled token at
  ordinary pre-publication boundaries;
- the crash child sentinel removes the previous fallback-panic false-pass.

Still blocking:

- incremental replacement still publishes the staged DB with its incomplete
  marker, then clears the marker on the live DB afterward. The matrix explicitly
  accepts the forbidden `old complete → None → new complete` state.
- MarkerCompletion is still post-replacement/after-PONR. It must complete and
  validate on staged storage, observe cancellation, then allow
  DatabaseReplacement.
- the crash hook aborts after backup but before `restore` begins, proving only
  that the untouched old DB remains. It does not interrupt the vulnerable
  mutation/restore point or prove automatic recovery from a partially modified
  live DB.
- full incomplete-run recovery can require marker completion, but the matrix
  skips MarkerCompletion for every Full mode.

Do not commit or merge this snapshot yet. `git diff --check` is clean; no Cargo
was run by the reviewer.

Issue #967, opened 2026-07-12
Harden managed CLI publication under cross-process contention


TheGreenCedar:
Parent: #921

## Scope

Harden the existing managed CLI provisioning path into a cross-process,
single-flight, staged and self-healing publication contract.

The current `provisionManagedCli` path already has a lock, temporary extraction,
checksum verification, version probing, and rename-based publication. This child
must close the failure/recovery gaps in that existing path rather than replacing
it with another installer abstraction.

## Acceptance criteria

- Concurrent processes requesting the same version elect one publisher; waiters
  reuse the fully verified result and never observe a partial version directory.
- Lock ownership records process/start identity, expires/reclaims abandoned
  owners safely, and cannot delete a live owner’s lock.
- Download and extraction occur outside the final version directory; archive
  hash, executable hash, executable version, and manifest content are verified
  before one atomic publication step.
- A malformed, partial, wrong-version, or checksum-mismatched installed version
  is quarantined with bounded retention and reprovisioned once from trusted
  release metadata.
- Publication remains fail-closed when quarantine or atomic replacement cannot
  be completed; no PATH or stale adjacent version is silently substituted.
- Native archive extraction remains supported on Windows, Linux, and macOS
  without introducing a new production dependency.
- Status/bootstrap diagnostics identify publisher, waiter, reclaimed lock,
  quarantine, reprovision, and terminal failure states without exposing secrets.
- Update the changelog and nearest operator/plugin documentation.

## Verification

- Real two-process same-version contention test with a delayed fake downloader;
  assert one fetch/extract/probe and identical verified result for both callers.
- Crash-after-lock and crash-before-publication recovery tests.
- Corrupt installed version quarantine/reprovision tests, including malformed
  manifest, wrong CLI version, checksum mismatch, and locked Windows executable.
- Atomic visibility test: a polling reader sees absent or complete, never
  partial publication.
- Existing plugin static suite, workflow policy, docs links, and diff check.
- Packaged-platform CI on Windows x64/ARM64, Linux x64/ARM64, and macOS x64/ARM64.

This child does not close #921 until full published-asset and managed-plugin
qualification is complete.

Refs #884
Refs #899

TheGreenCedar:
# Review #967

- Waiter timeout is shorter than the publisher's full bounded download retry window.
- Publication verifies `--version` but not MCP stdio protocol compatibility.
- External `tar` remains required and tests skip when it is absent.
[...]

Issue #921, opened 2026-07-10
Qualify every supported platform and harden managed runtime installation


TheGreenCedar:
Parent: #899

## Finding

Managed provisioning writes directly into final version directories, lacks a cross-process install lock, depends on external `tar`, does not recover a corrupt exact version, and ignores observed binary-version mismatches. Release execution proof is concentrated on Linux x64.

## Evidence

- `plugins/codestory/scripts/codestory-mcp.cjs:432`
- `plugins/codestory/scripts/codestory-mcp.cjs:572`
- `plugins/codestory/scripts/codestory-mcp.cjs:583`
- `plugins/codestory/scripts/codestory-mcp.cjs:708`
- Open Apple Silicon issue #887

## Acceptance Criteria

- [x] Provision each version under a cross-process single-flight lock.
- [x] Stage, checksum, probe, and verify observed CLI/protocol version before atomic directory publication.
- [x] Quarantine and reprovision a corrupt exact version once.
- [x] Use package-owned/native extraction with non-skippable platform tests.
- [x] Expose diagnostic MCP immediately while provisioning continues in the background.
- [ ] Execute install/update/ground/converge/cleanup on every supported OS/architecture.
- [ ] Document minimum Linux/glibc and explicit unsupported targets.
- [x] Post-publish proof executes every shipped asset.

## Verification

- Concurrent install and crash recovery tests
- Corrupt manifest/binary/archive tests
- Windows, Linux x64/ARM64, macOS ARM64, and declared macOS x64 matrix
- Fresh marketplace-managed plugin proof

## Non-Goals

- Do not infer compatibility from archive hashing alone.
- Do not close #887 without Apple Silicon live endpoint-survival evidence.

## Metadata

- Release: v0.15.0
- Area: CI/release/installer
- Priority: P1
- Wave: Next
- Refs #887

## Workstreams

- [x] #967 — harden managed CLI publication under cross-process contention
- [x] #1022 — serve diagnostic MCP during managed provisioning
- [x] #1030 — qualify managed lifecycle on all six shipped platforms
- [ ] Remaining — prove prior-version upgrade convergence, minimum Linux/glibc baseline, and #887 Apple Silicon Metal endpoint survival

Issue #899, opened 2026-07-10
Epic: trustworthy grounding and retrieval from v0.14.3 through v0.15.0


TheGreenCedar:
## Context

A full multi-lens review of CodeStory 0.14.2 found that the crate-level direction is worth preserving, but evidence integrity, incremental publication, MCP convergence, retrieval lifecycle, cross-platform proof, and duplicated product infrastructure need coordinated work.

Live managed-plugin evidence on 2026-07-10:

- local grounding is ready on managed/source 0.14.2;
- packet/search are blocked by stale sidecar state;
- the supported MCP repair entered `starting` and became abandoned;
- the standard parallel workspace test run exposed shared machine-state races, while the serialized workspace suite passed;
- production semantic documents include benchmark-shaped domain aliases, invalidating affected retrieval evidence.

This epic succeeds the completed roadmap in #501. It does not reopen that closed coordinator lane.

## Product outcome

CodeStory always provides the best safe grounding available. Compatible local navigation is not disabled merely because an update or sidecar rebuild is pending. When agent retrieval policy is enabled, startup and grounding converge it automatically without CLI repair commands, host restarts, profile selection, or run-ID management.

Indexes and sidecars are atomically published, project-isolated, recoverable after interruption, bounded in resource use, cross-platform qualified, and evaluated without production benchmark leakage.

### v0.14.3 - Trust & Recovery

Immediate stabilization release. Remove contaminated evidence, repair the current managed-plugin failure paths, introduce recovery guards, stop unbounded growth, isolate tests, and prove the packaged plugin.

Milestone: https://github.com/TheGreenCedar/CodeStory/milestone/10

### v0.15.0 - Grounding Runtime Rewrite

Architectural release. Replace the underlying indexing publication, convergence, project identity, sidecar lifecycle, scheduling, lexical retrieval, and duplicated evidence systems.

Milestone: https://github.com/TheGreenCedar/CodeStory/milestone/11

## Existing issues retained

- #884 remains the durable readiness/resource broker architecture issue.
- #887 remains the Apple Silicon Metal endpoint-survival report and v0.14.3 live-proof blocker.
- #897 remains the product-level automatic grounding convergence contract.

### v0.14.3

- [ ] No benchmark-shaped semantic alias catalog remains in production.
- [ ] Compatible local grounding remains usable when a newer release is available.
- [ ] MCP repair cannot abandon itself during ownership handoff.
[...]

PR #980, opened 2026-07-12
harden managed cli publication

TheGreenCedar:
## Context

Multiple Codex tasks can start the managed plugin adapter concurrently. The
0.14.3 launcher could let independent processes provision, replace, or clean the
same version directory without one authoritative publication transaction, and
could trust a manifest without proving the exact requested CLI version.

Closes #967
Refs #921
Refs #899

## What Changed

- Added a cross-process, PID/start-identity-owned publication lock with atomic
  initialization ownership, bounded waiting, stale-owner recovery, and exact
  live-owner protection.
- Added bounded reclamation for killed waiters' pending-owner artifacts without
  following links or deleting young/live ownership records.
- Elect one publisher; waiters reuse only the checksum- and exact-version-
  verified result.
- Stage and probe the complete release before atomic visibility, then fail
  closed on target reappearance or version mismatch.
- Quarantine invalid published versions with bounded retention; preserve a
  locked Windows executable and surface terminal quarantine failure rather than
  hiding it with a partial replacement.
- Redact release URL credentials/query/fragment metadata and require HTTPS
  outside explicit loopback test transport.
- Document publication roles, recovery, terminal failures, and retention.

## How To Review

Review the lock protocol in this order: process identity, pending-owner cleanup,
initialization claim, owner publication, stale recovery, release. Then review
`verifyPublishedManagedCli`, quarantine/retention, provisioning, and
`resolveManagedCli`. Finish with the real two-process kill/crash and Windows
locked-executable tests; injected filesystem failures supplement but do not
replace those OS-level proofs.

## Verification

- Full plugin static suite: 58/58 passed
- Focused crash/owner lifecycle suite: 4/4 passed
- 64 young live pending owners: zero process-start identity probes; 82.9ms
- Real two-process publisher/waiter election and exact-version reuse
- Real killed initializer and killed waiter recovery with no orphan owner files
- Real Windows locked executable through `resolveManagedCli`
- `node --check` on launcher and plugin test modules
- `node .github/scripts/check-workflow-policy.mjs`
- `node .github/scripts/check-doc-links.mjs` (69 files, 279 links)
- `git diff --check origin/dev/codestory-next...HEAD`

## Risk

Managed CLI installation is intentionally more fail-closed: filesystems that
cannot provide the required atomic ownership primitives or process identity now
produce a terminal managed-runtime diagnostic instead of racing publication.
Cleanup is root-scoped, exact-name filtered, capped at 64 candidates per pass,
and protects live or young ambiguous owners. No screenshot-visible UI changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,20 @@
 - Retried packaged-proof temporary directory removal after transient executable
   locks while keeping persistent cleanup failures fail-closed, preventing a
   successful Windows managed-plugin handoff from failing during teardown.
+- Made managed CLI provisioning cross-process single-flight: one publisher now
+  downloads, extracts, checksum-validates, probes, and atomically publishes a
+  release while waiters reuse the verified result. Lock ownership is published
+  atomically with process-start identity, acquisition fails closed when that
+  identity cannot be established, and incomplete initialization is
+  reclaimed only after its owner exits. Orphaned pending-owner files are
+  removed with the same process-identity checks, deferring expensive identity
+  probes until the stale threshold while live and young malformed artifacts
+  remain protected; malformed, wrong-version, or
+  checksum-mismatched installs are quarantined with bounded retention and
+  reprovisioned, while locked or unmovable targets fail closed. Publication
+  states are exposed as redacted plugin-runtime warnings, and native packaged
+  proof now includes macOS x64 alongside the existing five release targets.
+
 - Extended the retrieval generalization guard from derived query literals to
   direct and split Rust string dependencies on eval/query corpus paths across
   production crates, including benchmark manifests, query catalogs, packet
--- a/plugins/codestory/README.md
+++ b/plugins/codestory/README.md
@@ -31,14 +31,18 @@ and stay up with diagnostic `codestory://status` when managed setup fails.
 Ambient `PATH` binaries are reported as diagnostics only; installed plugin
 runtime does not launch them.
 
-After a managed runtime passes checksum and `--version` verification, the
-adapter retains that active version plus one verified pending upgrade or
-rollback. New installs publish from a verified staging directory under a
-PID/token lock. Older version directories are removed best-effort; live Windows
-executables, active installs, links, malformed manifests, and concurrent runs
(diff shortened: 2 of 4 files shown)


TheGreenCedar:
Independent source review of exact head `8f5a85e7`: six-target asset
selection is fixed and packaged CI is green, but publication is still blocked.

- **P1 — a healthy publisher can outlive every waiter.** The publisher holds
  the lock while fetching `SHA256SUMS.txt` and the archive. Each file can consume
  three 60-second attempts plus retry delays (`codestory-mcp.cjs:22-24`,
  `:563-577`, `:1005-1006`), while waiters give up after five minutes (`:27`,
  `:985`). Two healthy worst-case fetches exceed that bound, so concurrent
  startup can still fail `managed_cli_publish_locked` instead of reusing the
  verified result.
- **P1 — staging proves only `--version`, not stdio compatibility.** The
  pre-publication probe at `:1130-1156` never starts `serve --stdio` or performs
  MCP `initialize`. A binary can therefore pass version verification, publish,
  and then fail the only runtime interface the plugin needs.
- **P1 — stale-initializer reclaim retains an ABA deletion race.**
  `reclaimStaleManagedCliInitialization` reads an owner and later renames the
  path without revalidating that the inode/token is still the observed stale
  artifact (`:767-778`, `:676-684`). If the stale owner releases and a live
  contender installs a new `.initializing` link between those operations, the
  reclaimer can rename/delete the live claim. Apply the same descriptor/inode
  identity discipline already used for pending-owner cleanup.
- **P2 — archive verification is skipped when external `tar` is absent.**
  Production extraction is exclusively `spawnSync('tar', ...)` (`:582-590`),
  and the new contention/recovery tests skip when `tar --version` fails
  (`plugin-static.test.mjs:2501-2503`, `:2576-2578`, `:2662-2664`,
  `:2746-2748`). This does not prove the issue's native extraction requirement
  on every packaged platform; either make the platform prerequisite explicit
  and fail release CI when absent, or use a native platform path that the tests
  always exercise.

Do not merge this head. Preserve the current single-flight design; close these
four bounded gaps without adding another installer abstraction.

Issue #975, opened 2026-07-12
Reopen published storage after cross-process refresh lock handoff


TheGreenCedar:
Parent: #910

## Finding

The cross-process local-refresh lock recheck reuses the inspect runtime opened
before lock acquisition. If another process publishes while this caller waits,
that runtime can keep observing the old SQLite generation, classify it as stale,
and publish a second generation instead of coalescing.

## Evidence

- PR #973 Linux `linux-contracts` run 29178828621 failed
  `two_stdio_processes_observe_only_complete_generations_during_real_refresh`.
- Writer status returned typed `cache_busy` because publication changed again
  during its retry.
- `wait_for_local_freshness` rechecks through the pre-lock `inspect_runtime`
  after acquiring `local-refresh.lock`.

## Scope

Reopen published storage after cross-process lock handoff so the winning
publication is the freshness authority before any second refresh starts.

## Acceptance criteria

- The post-lock recheck uses a newly opened inspect runtime/publication, not a
  storage handle created before waiting.
- Two stdio processes racing one stale repository publish exactly one new
  generation.
- Status and grounding observe only old-or-complete-new state and the initiating
  status call does not fail with a second-publication `cache_busy` error.
- Add a deterministic regression that exercises the lock handoff rather than
  relying only on scheduler timing.
- Preserve bounded fail-closed behavior for genuinely repeated publication
  changes.

## Verification

- Focused two-process stdio publication contract
- Concurrent publication suite on Linux, Windows, and macOS
- Workspace checks and `git diff --check`

Refs #946
Refs #884
Refs #899

TheGreenCedar:
The Linux full-suite reproduction disproved this issue's stale-handle premise.

`RuntimeContext::open_project_summary` already opens fresh storage on every
summary read, and lock acquisition returns Busy immediately rather than waiting.
The actual visible transition is one incremental publication exposing
`old complete → fenced/None → new complete`: the staged database is atomically
published while its incomplete marker remains, and the live marker is cleared
only after cache finalization. Status can therefore observe both edges and
correctly fail its bounded retry with the currently misleading “changed twice”
message.

PR #981's fresh-runtime change is ineffective and is being closed unmerged.
The root fix belongs in #955/#973: complete the marker in staged storage after
all validation/search work but before database replacement; remove the
post-publication live-marker clear; place the marker fault boundary before the
point of no return; and prove live remains old until one atomic new complete
generation appears.

[...]

Issue #946, opened 2026-07-11
Serve the last complete publication during concurrent refresh


TheGreenCedar:
## Problem

CodeStory 0.14.3 promises project-scoped, multi-thread use, but one Codex thread refreshing a repository can block another thread's read-only MCP calls. On 2026-07-11, a second thread calling `status` for the same project first received `cache_busy`; later calls exposed `local_refresh.state=refreshing` while every local navigation surface remained blocked.

The writer must stay single-flight, but readers should not be coupled to the mutable refresh transaction. They should continue against the last complete publication and receive explicit refresh metadata.

Refs #897, #899, #910.

## Acceptance criteria

- [ ] Two independent MCP clients can use the same project concurrently.
- [ ] While client A performs a local incremental or full refresh, client B can call `status`, `ground`, and eligible graph/read tools against the last complete publication.
- [ ] Read-only calls do not fail with `cache_busy` merely because a refresh writer is active.
- [ ] Responses served during refresh identify the published generation and expose the active refresh owner/phase without presenting staged state as live.
- [ ] Concurrent refresh requests remain single-flight/coalesced and never create competing SQLite writers.
- [ ] A process-level concurrency regression starts two clients, holds one refresh across the publication boundary, and proves the other client observes only the old or new complete generation.
- [ ] Windows, Linux, and macOS CI exercise the concurrency contract or record an explicit platform qualification gap.

## Observed evidence

- Managed plugin and CLI: 0.14.3.
- Same explicit project was passed on every request.
- First concurrent `status` response: `cache_busy: CodeStory cache is busy or locked`.
- Subsequent status: local refresh active, index stale due an incomplete previous incremental run, and all local navigation surfaces disallowed.
- Agent packet/search was independently unavailable because the retrieval manifest was missing; that is a separate readiness lane and does not explain the local read lock failure.

Issue #907, opened 2026-07-10
Isolate integration tests from user caches and global broker state


TheGreenCedar:
Parent: #899

## Finding

Integration tests use temporary repositories and SQLite caches but still write sidecar namespaces, machine locks, and port allocations under the real user cache. Parallel readiness tests race through the same global native-runtime lock.

## Evidence

- `crates/codestory-cli/tests/stdio_protocol_contracts.rs:274`
- `crates/codestory-cli/tests/retrieval_bootstrap_contracts.rs:8`
- Default workspace run failed two broker observation tests; isolated and serialized runs passed

## Acceptance Criteria

- [ ] Every integration test process receives isolated sidecar, broker, port-registry, managed-runtime, and install identity roots.
- [ ] Tests never mutate the real user cache.
- [ ] Broker observation tests use injected machine-state roots.
- [ ] The default parallel workspace suite passes reliably.
- [ ] Add an invariant test that the real user cache is unchanged.

## Verification

- `cargo test --workspace --all-features` with default test concurrency
- Repeat the previously failing broker tests together
- Before/after user-cache inventory assertion

## Non-Goals

- Do not serialize the entire test suite as the permanent fix.
- Do not clean the real user cache from tests.

## Metadata

- Release: v0.14.3
- Area: Test isolation
- Priority: P1
- Wave: Now

TheGreenCedar:
Implemented and merged in #932 (`888baf6ff47c3ac77ba347e5526e1bda3e4b3687`).

What is now true:

- every integration-test CLI process resolves cache, stdio cache, install identity, plugin data, managed runtime, and sidecar registry state under one explicit process-and-test-thread root;
- direct `CARGO_BIN_EXE_codestory-cli` access outside the shared helper is rejected by a regression invariant;
- broker tests inject their machine-state root and propagate it across worker threads without changing feature-enabled production thread behavior;
- stdio repair fixtures and child servers use the same isolated root;
- a controlled decoy user profile remains unchanged across CLI retrieval status, while state is proven under the injected root;
- the default-concurrency `cargo test --workspace --all-features` gate passes.

[...]

Issue #884, opened 2026-07-08
Make CodeStory readiness broker a durable cross-process resource manager


TheGreenCedar:
## Context

CodeStory readiness state is currently split across MCP status handling, sidecar setup policy, repair commands, sidecar state files, and CLI readiness/progress code. That split leaks into UX: agents see command walls, local graph readiness is conflated with deep search readiness, and repeated status reads can start background repair work.

Live dogfood evidence from July 8, 2026:

- MCP runtime: managed CodeStory `0.13.12` for `C:/Users/alber/source/repos/codestory`.
- Local graph/navigation: fresh and usable.
- Agent packet/search: blocked with `retrieval_manifest_missing` and `retrieval_mode=unavailable`.
- `codestory://status` started another background sidecar repair PID while only inspecting status.
- Requested accelerator: `vulkan:Vulkan0`; status reports accelerated device detection, but there is no pre-repair GPU smoke gate proving the long rebuild will use the GPU meaningfully.

Related but narrower work: #800, #802, #855, #864.

## Problem

The broker cannot be just a shared Rust helper. Multiple MCP servers, CLI runs, plugin hooks, repos, and agents can run at once. CodeStory needs a durable runtime authority that owns readiness, leases, machine resources, process cleanup, and UX projections across processes and projects.

Without this, CodeStory can keep producing these failure modes:

- status-triggered repair children without a single durable owner
- stale or orphaned sidecar/process state before new work starts
- project/global state bleed between MCP sessions
- long sidecar rebuilds without proving GPU/native embedding viability first
- CLI and MCP reporting different readiness truth
- local graph blocked or visually degraded by unrelated deep-search sidecar failure

## Scope

Build a durable, cross-process, multi-tenant readiness/resource broker. File-backed and OS-atomic first; daemon only if file locks plus snapshots prove insufficient.

Broker operation identity must include:

```text
install_id
project_id / canonical_root_hash
workspace_root
profile: local | agent
run_id / agent_id
operation_kind
schema_version
cli_version
```

## Required Architecture

```mermaid
flowchart TD
  A[MCP server] --> B[Durable readiness broker]
  C[CLI command] --> B
  D[Plugin hooks] --> B
  E[Other agents and repos] --> B

  B --> F[Project scoped locks]
  B --> G[Agent sidecar locks]
  B --> H[Machine GPU/native queue]
  B --> I[Operation snapshots]
  B --> J[Process cleanup]
  B --> K[Debug telemetry]

[...]

PR #981, opened 2026-07-12
reopen after refresh lock handoff

TheGreenCedar:
## Context

PR #973 reproduced the same Linux stdio contract failure twice: while one
process completed a local refresh, another process reused an inspect runtime
opened before lock acquisition and could classify the old SQLite generation as
stale. Status also assembled its summary through separate autocommit reads, so
an in-place SQLite restore could cross generations during one response.

Closes #975
Refs #946
Refs #910
Refs #884
Refs #899

## What Changed

- Reopen the inspect runtime after acquiring `local-refresh.lock` before
  deciding whether another incremental refresh is necessary.
- Read every project summary inside one SQLite read transaction, preventing a
  response from combining rows from two in-place publication generations.
- Accept status only when the durable publication before and after the read is
  unchanged and matches the publication serialized in the response; retry up
  to three coherent read attempts.
- Include durable generation identity in the status cache key and cache only
  when the full key is stable across the read.
- Strengthen the real two-process contract so the initiating status succeeds
  and the final generation remains exactly old + 1 after both clients finish.

## How To Review

Review `wait_for_local_freshness`, then `StorageReadSnapshot` and
`open_project_summary_with_storage_inner`, and finally
`read_stdio_status_resource_cached`. The transaction pins related SQLite reads
to one coherent generation; the surrounding publication checks ensure that
only the current complete generation is returned and cached.

## Verification

- `two_stdio_processes_observe_only_complete_generations_during_real_refresh`
  passed after reproducing the pre-fix race with exact generation diagnostics.
- Durable publication match and cache-fingerprint unit tests passed.
- `cargo check -p codestory-store -p codestory-runtime -p codestory-cli` passed.
- `cargo clippy -p codestory-store -p codestory-runtime -p codestory-cli --all-targets -- -D warnings` passed.
- `cargo fmt --all` and `git diff --check` passed.

Running the complete stdio contract binary locally also exposed five separate
parallel-test isolation/worker-lifetime failures outside this diff. Those are
not claimed as passing here and are tracked in reopened issue #907 rather than
being serialized or weakened.

## Risk

Project-summary reads now hold a read transaction while gathering storage and
freshness state. SQLite WAL readers should not block publication, and the real
cross-process refresh contract exercises that path. Under sustained publication
churn, status still fails explicitly with `cache_busy` after the bounded retry
attempts instead of returning mixed state. No screenshot-visible UI changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,10 @@
 
 - Removed a plugin static test that froze exact grounding documentation copy;
   documentation remains covered by the repository link and diff checks.
+- Reopened published storage after acquiring the cross-process local-refresh
+  lock and served each status read from one complete immutable publication,
+  preventing redundant generations and spurious `cache_busy` failures during
+  status/grounding races.
 - Corrected the Codex managed-platform guide to list the restored macOS x64 CLI
   while retaining its no-managed-accelerated-sidecar boundary.
 - Retried packaged-proof temporary directory removal after transient executable
--- a/crates/codestory-cli/src/main.rs
+++ b/crates/codestory-cli/src/main.rs
@@ -2475,7 +2475,11 @@ pub(crate) fn wait_for_local_freshness(
             return Ok((summary, Some(output)));
         }
     };
-    let summary = inspect_runtime.open_project_summary()?;
+    // Another process may have replaced the published database while this
+    // caller waited for the cross-process lock. Reopen it before deciding that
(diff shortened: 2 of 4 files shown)


TheGreenCedar:
Closing unmerged: CI disproved the stale-handle premise. The corrected root cause and required staged-marker fix are recorded on #975 and belong to #955/#973.

Issue #965, opened 2026-07-12
Qualify hooks against status-only grounding regressions


TheGreenCedar:
Parent: #962

## Scope

Qualify the hook router against representative repository tasks and prevent a
return to status-only compliance.

## Acceptance criteria

- Define a small prompt matrix covering orientation, symbol lookup, call flow,
  change impact, broad architecture, blocked deep retrieval, and non-repo work.
- Assert the selected route and suppression decision deterministically.
- Run an agent-facing drill that records which CodeStory surfaces were actually
  used and whether source fallback occurred before an allowed graph route.
- Compare the new router with the current generic hook; publish counts and exact
  failure cases without turning the drill corpus into production literals.
- Keep instrumentation diagnostic/test-only and avoid persistent user tracking.

Refs #899

TheGreenCedar:
PR #982 merged before the independent exact-head review was available. Reopening
because the merged qualification does not prove this issue's acceptance bar.

Remaining gaps:

- the old status-only baseline score is hard-coded from prompt presence rather
  than produced by an executable baseline adapter;
- the fixture contains prompts and expected substrings, not separately labeled,
  sourced observed MCP traces for orientation, symbol, flow, impact, and
  blocked-fallback drills;
- the test calls the classifier directly instead of the production hook process;
- required surfaces are scored by prose substrings, which can pass incidental or
  negated tool names and freezes renderer copy.

Keep the deterministic cross-host fixture work. Add an executable former-policy
baseline, exercise current routing through the actual hook process, record the
observed drill traces without claiming tests execute MCP, and score structured
categories/exact tool tokens.

Exact-head review:
https://github.com/TheGreenCedar/CodeStory/pull/982#issuecomment-4950045567

Refs #962
Refs #899

Issue #963, opened 2026-07-12
Route repository prompts to the right CodeStory graph surface


TheGreenCedar:
Parent: #962

## Scope

Replace generic prompt repetition with a compact, task-aware grounding router
on the existing SessionStart and UserPromptSubmit hooks.

## Acceptance criteria

- Route orientation to `ground/files`, symbols to
  `symbol/definition`, call flow to `callers/callees/trail`, review/change work
  to `affected` with explicit git paths plus focused graph inspection, and broad
  questions to `packet` only when allowed/full.
- When packet/search is blocked, direct agents to allowed local graph surfaces
  before source fallback.
- Do not echo prompt text or inline the full grounding skill.
- Emit only for repository/source prompts; suppress general and coordination-only
  prompts.
- Keep startup/resume/compact reinjection, but remove dead heartbeat behavior
  unless the event is actually registered.
- Rename misleading packet preparation status copy.
- Add focused hook behavior/static tests.

Refs #899

TheGreenCedar:
# Review #963 / PR #971

- Repeated compact/resume lifecycle events are suppressed after the first occurrence in a task.
- Generic words such as `method`, `call`, `build`, `source`, and `architecture` inject CodeStory into non-repository conversation.
- The 900-character cap emits up to 938 characters and can remove the final evidence-boundary warning.
- Land #964 first, then rebase, because current `dev` skill guidance contradicts the router's local-graph fallback.

Exact-head review of `0d5cf05c`; no mutations performed.

Issue #962, opened 2026-07-12
Initiative: Route grounding hooks into real CodeStory graph use


TheGreenCedar:
Parent: #899

## Outcome

Make CodeStory hooks route agents into the smallest useful graph surface instead
of repeatedly enforcing a status call that often ends in ordinary source reads.

## Evidence

- The production activation path calls only `eventHeader`; the richer
  `getCodeStoryInstructions` task router/skill body has no production call edge.
- `UserPromptSubmit` repeats a generic startup path for every distinct prompt,
  echoes prompt text, and can truncate before useful routing guidance.
- The hook says “Preparing CodeStory packet...” although it creates no packet.
- When packet/search is blocked, the hook does not direct agents through the
  still-available local graph surfaces (`ground`, `symbol`, `definition`,
  `callers`, `callees`, `trail`, `affected`).

## Principles

- Hooks route; MCP tools provide evidence.
- Status is once per repository state, not ceremony before every file read.
- Prefer allowed local graph surfaces before source fallback.
- Repair sidecars only when the task genuinely requires packet/search.
- Do not inject stale packets, hidden source claims, or the full skill body.
- Do not add lifecycle hooks until the existing SessionStart/UserPromptSubmit
  surfaces are proven insufficient.

## Acceptance criteria

- Session startup injects a compact, truthful CodeStory task router and names
  the `codestory-grounding` skill without inlining it.
- Prompt submission emits only for repository/source work and selects a compact
  route for orientation, symbol ownership, call flow, review/change impact, or
  broad questions.
- Prompt text is not redundantly echoed and hook output cannot truncate in the
  middle of its operative contract.
- Agents are told when to reuse status, when to refresh it, and when not to
  trigger sidecar repair.
- Status/skill guidance makes the allowed local fallback explicit when deep
  retrieval is unavailable.
- Hook UX copy describes what the hook actually does.
- Static and behavior tests prove routing, deduplication, compact/resume
  reinjection, non-repo suppression, and no-op/blocked-surface behavior.
- A qualification drill demonstrates materially higher use of graph surfaces
  than status-only compliance across representative repository prompts.

## Workstreams

- [x] #963 - route repository prompts to the right graph surface
- [x] #964 - align skill/status guidance with local graph fallback
- [x] #965 - qualify against status-only grounding regressions

## Non-goals

- No automatic packet generation from lifecycle hooks.
- No automatic sidecar repair.
[...]

PR #982, opened 2026-07-12
qualify hook task routing

TheGreenCedar:
## Context

The task router shipped by #963 replaces generic status-first hook guidance, but
it needs qualification evidence that separates emitted instructions from graph
tools an agent actually used.

Closes #965
Refs #962
Refs #899

## What Changed

- Consolidated the existing inline route matrix into one 13-case test fixture
  covering route selection, instructed graph surfaces, and non-repo suppression.
- Added a small executable test-only adapter for the former generic
  UserPromptSubmit policy instead of constructing a `status` result inside the
  assertion.
- Emit per-case current/baseline route, suppression, instructed surfaces, and
  exact failures as the test diagnostic.
- Recorded five representative parent-saga MCP drills separately as source
  evidence: orientation, symbol ownership, call flow, explicit-path change
  impact, and blocked broad fallback.
- Kept the adapter, prompts, scoring, and drill trace test-only.

## How To Review

1. Inspect `hook-generic-baseline.cjs` as the executable semantic model of the
   old prompt-agnostic policy: it emits generic grounding with `status` for
   every prompt and suppresses none.
2. Inspect `hook-route-qualification.json` for expected emitted routes and the
   five explicitly sourced live drill traces.
3. Inspect the consolidated plugin test: route scoring uses real hook process
   output; the MCP trace is recorded evidence, not fabricated tool execution.

## Verification

- `node --test plugins/codestory/tests/plugin-static.test.mjs` — 50/50 passed.
- Current router — 13/13 cases had no route, suppression, required-surface, or
  forbidden-surface failures.
- Executable generic baseline — 0/13 cases qualified; the diagnostic emits the
  exact per-case reason (`route`, `suppression`, or missing instructed surface).
- Recorded live drills — orientation used `status -> ground -> files`; symbol
  ownership used `status -> symbol -> definition`; call flow used `status ->
  symbol -> callers -> callees`; change impact used `status -> affected` with
  both explicit hook/test paths matched; blocked broad used `status -> ground ->
  symbol -> trace -> source`. Status was reused, packet/search were blocked by
  `retrieval_manifest_missing`, and no repair was attempted.
- `node .github/scripts/check-doc-links.mjs` — 69 files / 279 links passed.
- `node .github/scripts/check-workflow-policy.mjs` — passed.
- `git diff --check` — passed.

The deterministic test does not claim to execute MCP graph calls. It qualifies
the hook's emitted route and instructed surfaces, then validates the shape of
the separately recorded live agent evidence.

## Risk

Low. The change is test-only plus changelog and replaces a duplicate inline
matrix. It freezes route categories and tool names, not full hook prose. No
screenshot-visible UI changed, so visual evidence is not applicable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,10 @@
   retrieval tasks through allowed project-scoped local graph surfaces, and
   repairs sidecars only when the task actually requires the blocked surface.
   Local navigation remains explicitly weaker than full packet/search proof.
+- Added a test-only hook qualification matrix for orientation, ownership, call
+  flow, change impact, broad retrieval, blocked-deep fallback, and non-repo
+  suppression, including an executable comparison with the former generic
+  status-first behavior and representative recorded agent drill traces.
 - Replaced generic status-only lifecycle guidance with compact repository-task
   routing for orientation, symbol ownership, call flow, change impact, and broad
   questions. Prompt hooks no longer echo user text, suppress non-repository
--- a/plugins/codestory/tests/fixtures/hook-generic-baseline.cjs
+++ b/plugins/codestory/tests/fixtures/hook-generic-baseline.cjs
@@ -1,0 +1,16 @@
+// Test-only executable model of the pre-router UserPromptSubmit policy: every
+// prompt received the same grounding instruction and none were suppressed.
+function emitGenericUserPromptPolicy() {
+  return {
+    hookSpecificOutput: {
+      additionalContext: [
+        'CODESTORY REQUEST GROUNDING ACTIVE',
+        '',
+        'Use CodeStory before source files for this user prompt.',
+        'Call status with the target repository before manual source reads.',
+      ].join('\n'),
+    },
+  };
+}
+
+module.exports = { emitGenericUserPromptPolicy };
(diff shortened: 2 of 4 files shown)


TheGreenCedar:
Independent review of exact head `b010fa2a`: qualification is not yet proven.

- **P1 — the status-only baseline is hard-coded, not executed.** At
  `plugins/codestory/tests/plugin-static.test.mjs:2693-2709`, `baseline` is
  assigned `"status"` whenever a fixture prompt is nonempty. All prompts are
  nonempty, so the claimed 0/13 score is guaranteed and cannot detect baseline
  behavior drift. Add a test-only semantic adapter for the former
  prompt-agnostic hook and score its actual output.
- **P1 — the required qualification drill is absent.** The fixture contains
  prompts and expected substrings only; there are no separately labeled,
  sourced observed MCP traces for orientation, symbol, flow, impact, and
  blocked-fallback drills. The test calls `routeForPrompt` directly, proving
  classifier text rather than the production hook process or materially higher
  graph-surface use. Run current cases through the actual hook process and keep
  observed tool traces as evidence without claiming the test executes MCP.
- **P2 — surface scoring can false-pass on prose.** `startsWith` freezes route
  labels and `includes(surface)` accepts incidental or negated words (for
  example “do not use trace”). Score a structured category and exact extracted
  instructed tool tokens; leave renderer prose free to change.

Cross-host fixture paths and deterministic iteration look correct. Do not merge
this head until the baseline and current drill evidence are executable rather
than asserted.

Issue #1067, opened 2026-07-13
make windows native embedding start identity reliable


TheGreenCedar:
## Context

PR #1052 owns the bounded cold-query budget change for #1039. Its retained Windows x64 package at exact head `701fa87e43e3a6de0938784a85a92934a2a27af5` launched the managed llama.cpp server on an AMD Radeon RX 7900 XT, selected `Vulkan0`, and offloaded 13/13 layers to the GPU. Readiness then failed before retrieval indexing or packet/drill execution with:

`native embedding identity did not converge after spawn: live native embedding process start identity is unavailable`

A second bounded retry failed at the same layer with the same error. PR #1052 has since advanced to `b81ef76d09a8333c5f3f46cb2483613dc6cf566e`, but that commit only trims scheduler proof code and does not change the Windows process-start identity path.

The Windows implementation currently shells out to PowerShell/CIM for each exact start-identity read. A transient missing CIM result makes an otherwise live, GPU-backed child fail closed during the before/snapshot/after ownership handoff.

## Acceptance

- Windows exact native-process start identity is read through the Win32 process handle and `GetProcessTimes`, without a PowerShell/CIM dependency for that field.
- The persisted `windows:<ticks>` identity format remains compatible with existing state.
- Missing processes return not-running while access/query failures remain fail-closed errors.
- Existing before/snapshot/after PID-reuse checks remain intact.
- One Windows regression test covers tick conversion and stable immediate reads from a spawned child.
- Focused retrieval, readiness-broker, check, clippy, and exact-head CI lanes pass.
- A packaged Windows x64 run reaches full retrieval with verified Vulkan proof and leaves no proof-owned process, container, listener, or active broker operation after cleanup.

## Boundaries

- Do not change #1052's query-budget behavior.
- Do not absorb #1059's project-identity V3 migration.
- Do not treat the packaged CodeStory MCP schema mismatch as part of this fix.

Refs #1039
Refs #1052
Refs #1048

PR #1059, opened 2026-07-13
migrate sidecar identity to v3

TheGreenCedar:
## Outcome

CodeStory sidecar state now has one project identity contract across Local, Agent, broker, retrieval, and packaged-proof surfaces. Current state uses schema 3; schema-2 state remains visible for migration inventory but is never reused, trusted, or destructively cleaned as a current runtime.

Closes #1034
Refs #1032
Refs #913
Refs #899

## What changed

- Versioned Local and Agent namespaces, state filenames, project/workspace IDs, and generation artifact scopes under identity schema 3.
- Bound persisted embedding selection to a redacted endpoint origin plus an install-keyed HMAC fingerprint. External endpoints cannot supply managed GPU proof.
- Bound managed native reuse, trust, and stop decisions to exact executable identity, argv, endpoint/port, and process-start identity. A live process without an exact persisted start identity is unverified; a dead PID remains distinguishable so stale ownership can be reaped.
- Kept legacy schema-2 state inventory-only while reconciling an in-flight compatible worker's terminal repair result into current status during migration.
- Preserved the simplified packaged-proof harness from #1073. The final proof delta against current dev is limited to identity-v3 helpers and live assertions, including the stable Darwin `/bin/sleep` exact-identity probe; the deleted speculative fake CLI/cleanup/status matrices were not restored.

## Final integration

- Current PR head: `4b8dd7439af0a780e868b7149c7e110530ae431c`
- Current dev base: `6c4e7d6aa482cc910c01cdd672d0efe943fa2929`
- Final base merge: `6ba684a3c2d728aa70ca6893985e6e39f9be8789`
- The merge retains #1058's managed-convergence fixes, #1072's Windows creation-time identity, #1073's proof simplification, and #1056's corpus boundary.

## Verification

Final integrated head:

- `cargo fmt --all -- --check` — passed
- `cargo check -p codestory-cli --locked` — passed
- `cargo test -p codestory-retrieval --lib --locked` — 362 passed, 2 ignored
- Focused CLI exact broker identity tests — passed
- Compatible legacy terminal-result reconciliation regression — passed
- Live-missing/dead-missing native start-identity regression — passed
- CLI sidecar-construction architecture contract — passed after routing the migration regression through the existing test-safe gateway
- `python3 .github/scripts/check-packaged-agent-proof.py --self-test` — passed
- `node .github/scripts/check-workflow-policy.mjs` — passed
- `git diff --check` — passed

[...]

PR #1052, opened 2026-07-13
raise cold sidecar query budget

TheGreenCedar:
## Outcome

Cold Agent-profile retrieval gets a bounded 1.5-second primary-query window. The packet-level deadline, stage order, cancellation, cache publication, and incomplete-result eligibility remain unchanged.

Closes #1039
Refs #998
Refs #918
Refs #1006
Refs #1063

## Why

The live Windows proof in #1006 reached a healthy full Agent generation with verified Vulkan work, but representative cold primary queries exhausted the former 1,000 ms cap before the unchanged packet/drill seam could complete. Warm requests passed on the same generation, isolating the remaining gap to the primary request window.

## What changed

- Raised the primary single-query default and cap from 1,000 ms to 1,500 ms.
- Preserved the existing reserve arithmetic, request deadline, cancellation traces, packet/search admission, ranker, and cache rules.
- Updated the changelog and the two documents that state the product query budget.
- Removed an intermediate CPU stats row and a test that replayed one machine's timing literals. The real Windows Vulkan cold proof remains the acceptance boundary.

## How to review

1. Confirm `sidecar_budget_ms` still clamps explicit requests and now defaults/caps at 1,500 ms.
2. Confirm the executor's later-stage reserve arithmetic is unchanged.
3. Confirm existing tests still cover the cap, remaining-budget allocation, and fail-closed deadline behavior.
4. Treat external Windows Vulkan evidence as required; this source change does not substitute synthetic timings for hardware proof.

## Verification

Current local head `b81ef76`:

- `cargo fmt --all -- --check`
- `cargo test --locked -p codestory-runtime sidecar_budget_respects_latency_cap -- --nocapture`
- `cargo test --locked -p codestory-retrieval graph_stage_uses_remaining_total_budget_across_query_orders -- --nocapture`
- `cargo test --locked -p codestory-retrieval non_broad_stage_deadline_remains_blocking_after_later_stages -- --nocapture`
- `node .github/scripts/check-doc-links.mjs` — 69 files, 281 links
- `git diff --check`
- Independent review: no P1/P2 findings

The branch still needs its final base refresh and exact-head hosted source proof after the Mac and runtime-identity lanes land.

## Hardware acceptance

The Windows orchestrator owns the remaining external proof. It must use the exact final packaged `windows-x64` artifact from an isolated cache and retain:

- provisioning, checksums, repair, and full status evidence;
- native Vulkan device, positive layer offload, verified meaningful work, and embed smoke;
[...]

Issue #1039, opened 2026-07-12
Make cold full-sidecar packet retrieval meet stage deadlines


TheGreenCedar:
Parent: #998

## Finding

The live #1006 packet/drill parity run proved a finalized Agent generation can report `retrieval_mode=full` while a cold representative packet is rejected because the required sidecar stages exceed their fixed deadlines. This is distinct from packet/drill adapter parity: the same unchanged parity gate passes after the exact-symbol path is warm.

## Exact evidence

Checkout: `261d09575a6b6ec7461587d973ef5d17c0ea8f71` plus the focused #1006 parity/runtime fixes.

Agent run: `drill-packet-parity`

Finalized generation: `1b25a35a2f8df931-32f2dc2c77b90afd`

Readiness: `profile=agent`, `retrieval_mode=full`, GPU proof `verified`, Vulkan0, meaningful accelerator work proven, embed smoke 14 ms.

Natural-language command executed by the existing `scripts/prove-drill-packet-parity.mjs` seam:

```text
packet --question "How does indexing feed packet retrieval?" --extra-probe WorkspaceIndexer --budget standard
```

Failure transcript:

```text
retrieval_unavailable: sidecar retrieval primary rejected query: sidecar retrieval trace `stage_deadline` is not eligible for primary results; sidecar_trace mode=full elapsed_ms=3154 candidates=0 resolved_hits=0 stages=[stage1_lexical added=0 elapsed_ms=574 cancel=stage_deadline; stage1b_qdrant_semantic added=0 elapsed_ms=0 cancel=zero_dense_anchors; stage2_scip_expand added=0 elapsed_ms=433 cancel=stage_deadline] top_candidates=[]
```

A cold exact-symbol attempt also failed:

```text
sidecar_trace mode=full elapsed_ms=4091 candidates=0 resolved_hits=0 stages=[stage0_scip_anchor added=0 elapsed_ms=632 cancel=stage_deadline; stage1_lexical added=0 elapsed_ms=84 cancel=stage_deadline; stage2_scip_expand added=0 elapsed_ms=202 cancel=stage_deadline; stage1b_qdrant_semantic added=0 elapsed_ms=0 cancel=zero_dense_anchors]
```

After filesystem/sidecar warming, the exact-symbol packet and unchanged parity gate passed against the same generation, with seven citations and one drill packet execution. That proves this is cold-stage scheduling/latency behavior, not missing artifacts or adapter divergence.

## Acceptance criteria

[...]

PR #1070, opened 2026-07-13
[codex] make windows native embedding start identity reliable

TheGreenCedar:
## Context

Windows native embedding startup records an exact process-start identity before the readiness broker accepts or reuses a managed llama-server process. The existing implementation obtained that identity through a fresh PowerShell/CIM query immediately after spawn. On the packaged Windows x64 Vulkan proof for #1052, that query failed to converge twice even though the process was live and offloaded all 13 layers to Vulkan.

This PR fixes the shared Windows identity source without changing #1052's scheduler work or #1059's identity-V3 migration.

Closes #1067
Refs #1039
Refs #1052

## What Changed

- Replaced the Windows PowerShell/CIM start-identity probe with `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` and `GetProcessTimes`.
- Owns the process handle with `std::os::windows::io::OwnedHandle` so cleanup remains RAII-managed and no dependency is added.
- Converts Windows `FILETIME` creation ticks to the existing .NET `DateTime.Ticks` epoch, truncates the sub-microsecond remainder exposed only by the native API, and preserves values previously serialized through CIM as `windows:<ticks>`.
- Returns `None` only when Windows reports that the PID no longer exists; process access and query failures remain fail-closed errors.
- Kept the existing before/snapshot/after PID-reuse checks and the CIM snapshot used for executable, arguments, and approximate start time.
- Added one Windows regression test for the known epoch conversion and repeated identity reads immediately after child spawn.
- Added an Unreleased changelog entry.

Exact implementation head: `6d3e3e2e5a319021fbd57a3579e4b7f6cae342bc`

## How To Review

1. In `crates/codestory-retrieval/src/sidecar.rs`, verify the Win32 declarations match the documented signatures and the process handle is adopted exactly once by `OwnedHandle`.
2. Verify `WINDOWS_DATETIME_TICKS_AT_FILETIME_EPOCH` preserves the old PowerShell `.CreationDate.ToUniversalTime().Ticks` value space and that the conversion removes only `FILETIME`'s sub-microsecond remainder.
3. Confirm that only `ERROR_INVALID_PARAMETER` maps to a missing process and every access/query failure is returned.
4. Confirm the existing process snapshot and double start-identity check are unchanged.
5. Run the Windows regression and focused identity/readiness lanes below.

## Verification

Passed locally on Windows x64 at `6d3e3e2e5a319021fbd57a3579e4b7f6cae342bc`:

- `cargo fmt --all -- --check`
- `cargo test --locked -p codestory-retrieval --lib windows_native_embedding_start_identity_is_stable_and_compatible`
- `cargo test --locked -p codestory-retrieval --lib native_embedding` (21 passed)
- `cargo test --locked -p codestory-cli readiness_broker` with an isolated test temp root outside `%USERPROFILE%` (43 relevant tests passed across unit and integration targets)
- `cargo check --locked -p codestory-retrieval -p codestory-cli`
- `cargo clippy --locked -p codestory-retrieval --lib -- -D warnings`
- `git diff --check`

The first readiness-broker run placed its injected root under the normal `%USERPROFILE%\AppData\Local\Temp` path and triggered the existing test's platform-cache prefix assertion. Rerunning with `TEMP` and `TMP` set to the isolated `C:\codestory-test-temp-1067` root passed; that root was removed afterward.

## Risk

- The change is Windows-only and preserves the public identity format, but the packaged Windows x64 Vulkan acceptance proof is still required before merge.
- Exact-head source CI, independent re-review, and the Windows x64/ARM64 package cells are pending.
- After merge, #1052 must rebase and rerun its unchanged-generation packet/drill parity proof; this PR does not claim that downstream acceptance yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
 
 ### Fixed
 
+- Replaced the Windows native embedding process-start identity CIM probe with
+  the native process creation time. Immediate post-spawn identity reads no
+  longer depend on CIM convergence and retain the existing serialized format.
 - Made incremental freshness compare verified parser source hashes when file
   mtimes match. Same-timestamp edits now schedule reindexing instead of leaving
   stale graph data current, while legacy and non-parser rows retain the
--- a/crates/codestory-retrieval/src/sidecar.rs
+++ b/crates/codestory-retrieval/src/sidecar.rs
@@ -802,6 +802,9 @@ fn normalized_identity_path(path: &str) -> String {
 
 #[cfg(windows)]
 pub fn native_embedding_process_start_identity(pid: u32) -> Result<Option<String>> {
+    if pid == 0 {
+        bail!("native embedding process pid must be greater than zero");
+    }
     let raw_handle = unsafe { OpenProcess(WINDOWS_PROCESS_QUERY_LIMITED_INFORMATION, 0, pid) };
     if raw_handle.is_null() {
         let error = io::Error::last_os_error();
@@ -836,7 +839,10 @@ pub fn native_embedding_process_start_identity(pid: u32) -> Result<Option<String
 fn windows_datetime_ticks_from_filetime(filetime: &WindowsFileTime) -> Result<u64> {
     let filetime_ticks =
         (u64::from(filetime.high_date_time) << 32) | u64::from(filetime.low_date_time);
-    filetime_ticks
+    // Win32_Process.CreationDate exposes microseconds, so discard sub-microsecond
+    // FILETIME ticks to preserve identities serialized by the previous CIM query.
+    let legacy_filetime_ticks = filetime_ticks / 10 * 10;
+    legacy_filetime_ticks
         .checked_add(WINDOWS_DATETIME_TICKS_AT_FILETIME_EPOCH)
         .context("convert Windows process creation time to DateTime ticks")
 }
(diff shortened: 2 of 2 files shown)


TheGreenCedar:
### Exact-head review context

Independent read-only review accepted exact head `6d3e3e2e5a319021fbd57a3579e4b7f6cae342bc` against base `1216d19ce4a07e5a4cfee2e6d18b817abc0e916d` for source-proof dispatch.

The first review of `a204c49402b30bec26d3942b65933eaeb1cd575b` correctly blocked promotion because raw `GetProcessTimes` values retain 100 ns precision while the replaced CIM `CreationDate` value is microsecond-granular. Same-PID probes demonstrated a 0-9 tick compatibility gap that could reject an identity persisted before upgrade. The current head truncates only that sub-microsecond remainder before applying the .NET epoch, and the regression now uses a known value with an 8-tick remainder so removing the compatibility conversion fails the test. The current head also fails closed for PID 0 instead of treating `ERROR_INVALID_PARAMETER` as a vanished child.

Review confirmed:

- `OpenProcess`/`GetProcessTimes` declarations and access rights match the Win32 contract.
- `OwnedHandle` adopts each successful process handle exactly once.
- Missing-process handling is distinct from access and query failures.
- The persisted `windows:<ticks>` value remains compatible with the old CIM writer.
- Existing before/snapshot/after PID-reuse checks and the CIM executable/argument snapshot are unchanged.
- Non-Windows implementations are unchanged.
- P1/P2 findings on the current head: none.

Local exact-head verification passed: formatting, the Windows regression, 21 focused native-embedding tests, the readiness-broker lane, retrieval/CLI checks, focused clippy with warnings denied, and `git diff --check`.

Disposition: apply `review-accepted` and dispatch exact-head source proof. Keep the PR draft pending source proof and full packaged platform proof, including Windows x64 and Windows ARM64.

TheGreenCedar:
### Windows packaged live proof

Exact-head source proof and the Windows package cells completed for `6d3e3e2e5a319021fbd57a3579e4b7f6cae342bc` in [platform run 29281875582](https://github.com/TheGreenCedar/CodeStory/actions/runs/29281875582). Both Windows x64 and Windows ARM64 built and passed their package smoke and managed-plugin handoff lanes. The overall dispatch is not green: the repo-scale stats job failed and the macOS package cells are still waiting.

I downloaded the exact Windows x64 artifact into a new isolated cache and verified the archive checksum:

- `codestory-cli-v0.14.3-windows-x64.zip`
- SHA-256 `662ed6f80f3025e475320c3a25affdd1157158d8549c29f2a757f4c34f030dd2`

The live `ready --goal agent --repair` retry still failed before indexing or packet execution:

`native embedding identity did not converge after spawn: live native embedding process start identity is unavailable`

This failure is narrower than the original blocker. The new native `OpenProcess`/`GetProcessTimes` identity capture succeeded immediately; otherwise the command would have failed with the distinct `unavailable immediately after spawn` error. The later convergence check still obtains `NativeEmbeddingProcessSnapshot.started_at_epoch_ms` through PowerShell/CIM. On this Windows host, `Win32_Process.CreationDate` is already a `System.DateTime`, so passing it to `ManagementDateTimeConverter::ToDateTime` fails and leaves the approximate start time unavailable.

The native log again proves AMD Radeon RX 7900 XT Vulkan execution with 13/13 layers offloaded and a live loopback listener. `gpu_proof=verified`, retrieval `full`, packet, and drill remain unproven because readiness stopped at identity convergence.

Cleanup is complete and proof-scoped: the managed PID is dead; proof-owned process, container, and listener counts are zero; the sidecar state file is absent; and the stale broker snapshot/coordination lock were preserved as output evidence and removed from the isolated cache. Unrelated existing llama-server and Qdrant resources were not touched.

The PR stays draft. The revised minimal fix is to keep CIM for executable path and command line but source the Windows snapshot's approximate start time from the same native creation `FILETIME`. That requires a new exact head, independent re-review, source proof, Windows x64/ARM64 package cells, and a fresh isolated packaged live proof.

Issue #1299, opened 2026-07-19
[P1] Stream canonical search symbols without rebuilding a projection table


TheGreenCedar:
Parent: #1275
Release: v0.16.0

## Context

The retained 94,411-file proof spent 19m20s rebuilding `search_symbol_projection` and 32m32s building the symbol index. Even after the single-writer Tantivy fix, current runtime still materializes all 8.55M `Node` rows, rewrites the same 8.55M IDs/names into a derived SQLite table, reads that table back, and holds another full projection vector while constructing runtime search state.

The canonical search label is already a pure node-table expression: nonempty `qualified_name`, otherwise `serialized_name`. The projection table is not an independent product truth. Retrieval and SCIP currently force a rebuild only because they consume that derived table.

CodeStory packaged grounding failed closed at CoreFreshness for the materialization audit (`project_unavailable`, `next_action=use_source_inspection`), so this plan uses exact-head source inspection. The retained corpus remains untouched until the combined parent proof.

## Contract

- add store-owned keyset readers that stream canonical search symbols directly from `node` in ascending `node_id` order, with a separate detail reader joining the owning file node for SCIP locations;
- each page is bounded by an explicit production batch size, advances by the last returned ID, and rejects a zero limit. No OFFSET scan and no whole-corpus store return value;
- runtime persisted-search construction opens one Tantivy symbol session, feeds canonical symbol pages into it, and incrementally fills only the runtime structures that must remain resident. It must not call `get_nodes`, rebuild `search_symbol_projection`, or collect a second full projection vector;
- completed-generation reuse and live rehydration stream the same canonical pages, validate streamed row count against the core node count and completion marker, and revalidate publication identity before exposing the result;
- cancellation is checked between pages and before commit/completion. A cancelled or failed build drops the writer, discards only the unpublished search generation, and leaves the previous complete core/search publication usable;
- retrieval sidecar preparation no longer creates or repairs `search_symbol_projection`; SCIP emission streams canonical detail pages directly from the core;
- preserve `search_symbol_projection` schema and public compatibility methods in this child, but remove it as a production source of truth. Do not pay a one-time multi-million-row migration delete on open;
- keep existing `search_projection_rebuild_ms` wire compatibility and report zero when no rebuild occurs. Add explicit optional symbol-stream duration, rows, and batch-count telemetry so the eliminated phase and bounded replacement are visible;
[...]

Issue #1275, opened 2026-07-18
Eliminate staged full-refresh transaction and commit barriers


TheGreenCedar:
Parent: #1237
Root: #1202
Release: v0.16.0

## Accepted baseline

The retained clean-corpus proof at exact integrated head `0fb25885d586c3c718e268932896673b9e240a33` completed successfully on the case-sensitive APFS sparse image:

- 94,411 parsed files; 65,532 stored files
- 8,552,335 nodes; 5,690,718 edges
- 16.6 GiB core SQLite database; roughly 41 GiB of search artifacts
- 14,243.46 seconds wall time (3h57m23s)
- quick check and doctor core passed; verified exclusion proof complete

Retained evidence: `/Volumes/CodeStoryLinuxCorpus/runs/issue-1237-0fb25885-rerun-FbYOJw/run-info.md`

Baseline phase attribution:

| Phase | Time | Wall share |
| --- | ---: | ---: |
| Symbol search index | 32m32s | 13.7% |
| Search projection rebuild | 19m20s | 8.1% |
| Summary snapshot | 14m30s | 6.1% |
| Edge resolution | 6m54s | 2.9% |
| Publication | 6m49s | 2.9% |
| Projection flush | 3m45s | 1.6% |
| Parsing | 1m46s | 0.7% |
| Unattributed | 2h28m53s | 62.7% |

## Retained stopped candidate

The later full-refresh attempt at exact CodeStory head `f447f934e936b98c750c72f29fab4c01fe261009` processed all 94,411 source files but was intentionally stopped after 12,142.20 seconds (3h22m22s), before final search construction or core publication.

- source ingestion completed in about 50m07s
- max RSS 4,434,591,744 bytes; peak footprint 11,673,627,232 bytes
- staged database 9.1 GiB; staged WAL 627 MiB; retained cache 9.7 GiB
- no `source-index.json`, no new core publication

Retained evidence: `/Volumes/CodeStoryLinuxCorpus/runs/issue-1275-final-f447f934-XiJmRj/`

That evidence is sufficient for the diagnosed endpoint-index defect and remaining bottleneck inventory. Do not resume the run or start another complete Linux corpus proof without explicit authorization.

## Integrated children

Current integration head is `a3321ada123ced39390682d2e737ed66102d21d8`, tree `99a355d8e7606784ff531522e08c6e9932b42113`.

- [x] #1287 — batch artifact-cache upserts once per projection chunk and expose write/transaction telemetry
[...]

PR #1300, opened 2026-07-19
[codex] Stream canonical search symbols without projection rebuilds

TheGreenCedar:
Closes #1299
Refs #1275

## Context

The retained v0.16 corpus proof spent 19m20s rebuilding `search_symbol_projection`, then runtime loaded all 8.55M complete `Node` rows before search construction. The derived projection contains only node ID plus the canonical label already available from `node`.

This change removes that materialization from product runtime and retrieval paths while keeping the legacy schema and public store methods compatible.

Exact candidate: `bfdd59c2c2a2842f43481318e55d321dc74346ab`

## What changed

- add store-owned canonical symbol and SCIP-detail keyset readers with explicit bounded limits, stable node-ID ordering, canonical qualified-name fallback, counts, and typed zero-limit rejection;
- stream 4,096-row pages through one Tantivy symbol session for new search generations, while incrementally filling only the resident name/fuzzy structures;
- stream the same source for completed-generation reuse, live rehydration, and pinned retrieval reads, with count and publication revalidation;
- stop retrieval preparation and SCIP emission from repairing or reading `search_symbol_projection`;
- preserve the legacy projection table and compatibility methods without a migration-time delete;
- retain `search_projection_rebuild_ms` as zero and add optional symbol stream time/row/batch telemetry to the DTO and CLI;
- add an architecture contract preventing product runtime/retrieval from returning to whole-node or legacy-projection search construction.

## Review focus

- `crates/codestory-store/src/storage_impl/mod.rs`: canonical SQL and page contract;
- `crates/codestory-runtime/src/lib.rs`: streaming writer, generation reuse/rehydration, cancellation, count validation, and timer boundaries;
- `crates/codestory-retrieval/src/{index,scip_index}.rs`: removal of legacy projection repair and canonical SCIP pages;
- compatibility: `SearchEngine::index_nodes` remains available under `benchmark-support`; legacy store projection APIs remain unchanged.

## Verification

- `cargo test -p codestory-store --lib --locked` — 131 passed
- `cargo test -p codestory-retrieval --lib --locked` — 194 passed, 3 ignored measurement/live-runtime lanes
- runtime persisted generation, full/incremental publication, missing/corrupt/count-mismatch, writer failure, cancellation, and atomic transition lanes — passed
- 4,100-row runtime fixture — two pages, exact rows/docs, one writer/commit/reload, cancellation after the first non-final page commits zero docs
- `cargo test -p codestory-contracts --lib --locked` — 38 passed
- `cargo test -p codestory-cli --test architecture_contracts --locked` — 13 passed
- CLI timing render test — passed
- `cargo check -p codestory-bench --bench tantivy_symbol_commits --locked` — passed
- focused multi-package check and all-target clippy with `-D warnings` — passed
- `node .github/scripts/check-doc-links.mjs` and `git diff --check` — passed
- independent review found two pre-commit issues (benchmark API gating and timer attribution); both were fixed and the narrow re-review found no remaining P0-P3 findings

Packaged CodeStory grounding failed closed at CoreFreshness because `crates/codestory-runtime/src/lib.rs` exceeds the source-size policy, so review used exact-head source inspection as directed.

## Risk and follow-up

The resident node-name and Nucleo structures remain intentionally in memory, and SCIP still buffers its final output artifact. No repo-scale run is claimed here. #1275 owns the single combined 94,411-file rerun and any speedup or peak-memory claim.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@
   document checkpoints and commit/reload once, instead of reopening and
   committing the writer for every 8,192 symbols. Index telemetry reports
   symbol documents, writers, commits, and reloads.
+- Runtime and retrieval now stream canonical search symbols directly from the
+  node table in 4,096-row keyset pages. Fresh full and incremental builds no
+  longer rewrite the legacy SQLite search projection or materialize every
+  complete graph node before search construction; telemetry reports stream
+  time, rows, and batches alongside the single Tantivy commit.
 - File-backed full refreshes now overlap preparation and Rayon parsing with a
   dedicated staged SQLite writer through a capacity-one queue. Cancellation
   drains only accepted chunks, incremental indexing keeps its serial path, and
--- a/crates/codestory-cli/src/main.rs
+++ b/crates/codestory-cli/src/main.rs
@@ -8949,6 +8949,9 @@ mod tests {
             full_refresh_chunk_planning_ms: Some(5),
             cache_refresh_ms: Some(6),
             search_projection_rebuild_ms: Some(61),
+            search_symbol_stream_ms: Some(60),
+            search_symbol_stream_rows: Some(8192),
+            search_symbol_stream_batches: Some(2),
             search_symbol_index_ms: Some(62),
             search_symbol_index_docs_written: Some(8192),
             search_symbol_index_writer_count: Some(1),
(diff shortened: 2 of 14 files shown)


TheGreenCedar:
Exact-head gate is complete for `4398fee43986310aa76e4f2b6e22af4151531399...bfdd59c2c2a2842f43481318e55d321dc74346ab`.

- independent review: no P0-P3 findings
- local HEAD, upstream branch, and PR head are identical
- exact tree: `a4b7bc102973c26a8cab6e4ab2529b33681cc1d3`
- all non-skipped hosted checks: success
- merge state: clean and mergeable
- non-final-page cancellation: zero committed Tantivy docs and no completion marker
- full and incremental publication fail/cancel matrices: passed with the new page boundary

The packaged grounding gap remains explicit: the changed runtime source exceeds the current source-size policy, so exact Git/source review and runtime proofs are the basis. No corpus rerun is claimed in this child.

Issue #1364, opened 2026-07-21
[P1] Fix Windows staged database durability sync


TheGreenCedar:
## Outcome

Make disposable full-refresh database sealing succeed on Windows by opening the staged SQLite database with the access required by `File::sync_all`.

## Reproduction

At frozen dev SHA `3d8e4b24b69f650f5cc872bd5fa301d0dcc66917`:

```text
cargo test --locked -p codestory-store storage_impl::tests::staged_promotion_abort_recovers_old_or_complete_new_and_cleans_artifacts -- --exact --nocapture
```

fails while seeding the sealed staged generation:

```text
Failed to sync staged database ...: Access is denied. (os error 5)
```

`seal_disposable_full_build` currently uses `File::open`, which supplies a read-only Windows handle before calling `sync_all`.

## Acceptance

- Open the staged database with the write access needed for the durability flush.
- Keep WAL checkpointing and parent-directory sync behavior unchanged.
- The exact Windows reproduction passes.
- Focused store tests and release policy checks remain green.

Refs #1189

Issue #1189, opened 2026-07-15
[v0.17 proof] Qualify the exact integrated head and installed runtime


TheGreenCedar:
Parent: #1179

## Outcome

Run the complete, ordered proof program against one independently accepted merge-ready SHA and the installed runtime built from it, then record only claims supported by that evidence.

## Current checkpoint

Draft promotion PR #1259 tracks `dev/codestory-next` at `5f00ef470dd1b09127320a2c9553241af6032e9b`, tree `34de081cd0356ac5503632eddefda5628a289a47`, 216 commits ahead and 0 behind `main` at `85a877e1f0bc9986d203b09efe70f2956fb5b32e`. This checkpoint is not frozen or accepted.

The installed-launcher child #1327 is closed. The structural-policy child #1331 and retry/retained-local-capability child #1333 are also closed after their accepted installed-host proof; they are not live blockers. #1338 is source-integrated and remains open for its exact installed-host architecture-ranking replay. #1197, #1202, #1196, #1200, #1221, #1213, and #1233 retain their respective acceptance contracts. Final source, package, protected-hardware, installed-runtime, and release-ledger proof has not run on an unchanged head.

## Dependencies

- All release-blocking children of #1179.
- All correctness and usefulness children required by the declarative claim graph.

## Consolidated final receipts

- **Windows:** one final exact-candidate Windows receipt owns package/archive smoke, physical Vulkan, isolated candidate-installed behavior, authenticated final-server exit within 120 seconds, removal of both temporary roots, and exact closeout. That one receipt is referenced by #1372, #1378, #1381, #1383, #1397, and the Windows portion of #1221. Those source-merged leaves stay open off-milestone until the receipt passes; they do not get separate qualification runs.
- **Axios:** #1395 is the single milestone-facing gate. #1403 is its only active source blocker and #1398 / PR #1402 is its only evidence lane. Failed run `29962963157` is retained; exactly one protected retry is allowed after serial source integration and exact-head review.
- **Other installed behavior:** the #1351 exact-path negative/positive replay, #1338/#1197 grounding matrix, and #1396 protected latency values are folded into the frozen-candidate receipts rather than spawning more milestone children.

## Acceptance criteria

- [ ] Pin one exact integrated SHA after independent review and exact-head re-review.
- [ ] Complete required source, concurrency/fault, package, native, protected hardware, workflow, and benchmark lanes on that SHA.
[...]

PR #1365, opened 2026-07-21
[codex] Fix Windows staged database durability sync

TheGreenCedar:
## Context

The exact v0.16 Windows repo-scale candidate failed while sealing its disposable full-refresh SQLite database with `Access is denied. (os error 5)`. The shared seal opened the file read-only and then called `sync_all`, which requires a write-capable Windows handle.

Closes #1364
Refs #1189

## What Changed

- Open the existing staged database with write access before its durability flush; creation, truncation, checkpointing, parent sync, and promotion order are unchanged.
- Give the abort sentinel the same valid Windows sync handle so the crash proof reaches `abort()` instead of passing after an earlier panic.
- Drop three live test connections before deleting their Windows fixtures.
- Record the release-significant Windows fix in the v0.16 changelog.

## How To Review

Start at `Storage::seal_disposable_full_build`: the only production change is the access mode of the temporary durability handle. Then inspect the test-only handle and connection lifetime changes.

## Verification

- Pre-fix exact reproduction: `snapshot_store::tests::disposable_full_refresh_seals_wal_before_promotion` failed with OS error 5.
- Post-fix: all 169 `codestory-store` library tests pass on Windows.
- Exact crash/promotion regressions pass:
  - `snapshot_store::tests::disposable_full_refresh_seals_wal_before_promotion`
  - `storage_impl::tests::process_abort_during_disposable_build_never_mutates_live`
  - `storage_impl::tests::staged_promotion_abort_recovers_old_or_complete_new_and_cleans_artifacts`
- `cargo fmt --all -- --check`
- `git diff --check`
- `python .github/scripts/check-codestory-release.py --version 0.16.0`
- `node .github/scripts/check-workflow-policy.mjs`
- Independent review: no findings.

## Risk

Low. The handle opens an already-created SQLite database without create or truncate. SQLite already owns the staged database read-write; the change only lets Windows flush that file before promotion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ CodeStory 0.16 makes repository search more accurate and reduces duplicate model
 - Retrieval indexing now binds the selected project runtime after compatibility
   preflight and before executing an automatic incremental refresh, preventing
   exact-head repo-scale evidence runs from failing with an unopened project.
+- Disposable full-refresh storage now opens the staged SQLite database with a
+  writable handle before its durability flush, so Windows can seal and promote
+  the candidate instead of failing `sync_all` with access denied.
 - A new explicit `retrieval republish-projections` writer can adopt the current
   semantic and dense-selection policy from one complete stored core without
   source discovery, parsing, or source-file reads. It validates the pinned
--- a/crates/codestory-store/src/storage_impl/mod.rs
+++ b/crates/codestory-store/src/storage_impl/mod.rs
@@ -5006,7 +5006,9 @@ impl Storage {
             )
         })?;
         let sync_started = Instant::now();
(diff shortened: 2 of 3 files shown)


TheGreenCedar:
Independent exact-head review accepted commit `d1e6b3da`: no blocking findings.

The production change only replaces the read-only durability handle with a write-capable, non-creating, non-truncating handle before `sync_all`. WAL checkpoint, file sync, parent sync, close, and promotion ordering remain intact. The test changes correct Windows handle lifetimes and ensure the abort proof reaches the intended process abort.

Windows verification: all 169 `codestory-store` library tests pass, including the exact seal, crash, and promotion regressions that failed or were invalid before the fix.

Issue #1381, opened 2026-07-22
Wait for the exact candidate server before Windows package cleanup


TheGreenCedar:
## Outcome

Make Windows package qualification wait for the exact proven per-user embedding-server process to exit normally before its temporary unpack directory is removed.

## Reproduction

Exact Windows-only run `29894398392`, job `88841433662`, passed the `windows-x64` archive/version smoke (`packaged-version-proof-windows-x64`, artifact `8519809936`). Both isolated MCP host transcripts in managed-proof artifact `8519828759` contain only successful calls: host A completed 11 and host B completed 7. The proof checker then failed while leaving its temporary package scope:

> `PermissionError: [WinError 5] Access is denied: codestory-cli.exe`

Job cleanup found the unpacked `codestory-cli.exe` still running. The per-user server intentionally remains alive through its bounded idle period after both hosts close, while the checker currently removes the temporary unpack directory immediately.

## Acceptance

- After successful isolated-host proof, wait only on the exact server process identity proven by the run.
- Allow the server to exit through its normal idle lifecycle; do not broad-kill, enumerate by name, or weaken orphan cleanup.
- Fail closed if process identity changes or the process does not exit within the bounded grace.
- Preserve both successful host transcripts and package provenance before cleanup.
- Focused checker self-tests cover the wait and identity boundary.
- A physical Windows regression proves the unpack directory can be removed and no candidate `codestory-cli.exe` remains.
- A fresh exact-head Windows-only coordinator run reaches managed handoff, protected Vulkan, isolated candidate-installed behavior, and closeout.

## Boundaries

This fix does not itself prove Vulkan, installed-runtime, Mac/Metal, answer quality, publication, promotion, or release readiness.

Refs #1372
Refs #1221
Refs #1189

TheGreenCedar:
Implementation checkpoint

Draft PR #1382 is open at exact head `315eb95794c8a84c9d5e49694eaa3aff40388205` / tree `4f40f3c568bcec7efc2bcc013f7153c016b9df5e` from base `f56e1a65d0ed59a4f0e864078b7c9ce7b98ff0ee`.

The checker now verifies the live candidate server image, opens a native handle bound to its PID and process-start identity while both hosts still attest it, closes the hosts, and waits for that exact process to leave through the frozen 60-second idle path plus the existing 2.5-second qualification grace. It never enumerates or kills by name and does not relax temporary-directory cleanup.

[...]

Issue #1372, opened 2026-07-21
[P1] Route Windows-only qualification without unrelated release evidence


TheGreenCedar:
## Context

Exact Windows platform run `29866189303` accepted frozen source proof for `d81bba4d49df9582b94e24a38f48e7a6284c9a65`, then queued `release-evidence / Measure release candidate` on the offline Mac-hosted Linux ARM64 runner. The online protected Windows/Vulkan runner could not be scheduled because frozen `scope=windows` currently requires that unrelated release-quality producer.

`scope=windows` is a bounded platform-proof route, not a statement that macOS is unavailable or unsupported. Mac/Metal remains part of the v0.16 qualification matrix and is coordinated in its own lane. Skipping unrelated release-quality work for this Windows-scoped route must not weaken or fabricate the full release ledger.

## Outcome

Give `scope=windows` the same bounded, truthful server-behavior lane already accepted for explicit Linux scope:

- skip protected ARM64 release-evidence;
- package the exact Windows candidate;
- run protected Vulkan and isolated candidate-installed Windows server behavior against the frozen calibration bundle;
- retain exact provenance and explicit answer-quality/release-readiness non-claims;
- keep `scope=full` and publication-capable release workflows dependent on exact release-quality evidence.

## Acceptance criteria

- [ ] Windows-only platform scope does not queue the Linux ARM64 release-evidence job.
- [ ] Windows-only package, protected Vulkan, and candidate-installed proof run with `server_behavior_only` and no quality artifact.
- [ ] Full/release scopes still require exact-head release evidence and still emit no lower-tier aggregate claim.
- [ ] Closeout requires release-evidence skipped for Windows-only scope and successful package/Vulkan jobs.
- [ ] Workflow policy and mutation tests reject removal of these boundaries.
- [ ] One live exact-head Windows-only coordinator run reaches green closeout without requiring the unrelated Mac-hosted release-evidence producer.

## Non-claims

This lane does not prove answer quality, the authenticated 12-cell pre-publish ledger, Mac/Metal, marketplace installation, post-publish behavior, promotion, or release readiness.

TheGreenCedar:
PR #1373 merged into `dev/codestory-next` only after exact-head review and green source checks.

- Frozen dev SHA: `a99b922c9d20cf13f960567f02d0e20286365fba`
- Frozen tree: `3c9fd0e30e6dc818a9a1aaae7f8119d78953beeb`
[...]

Issue #1221, opened 2026-07-16
[v0.17 proof] Qualify per-user embedding server ownership in installed runtimes


TheGreenCedar:
Parent: #1213

## Outcome

Qualify the per-user auto-spawned embedding server in exact packaged and
installed runtimes before closing the shared-admission parent.

This qualification runs in the same PR as #1220 after the source candidate and
its constants are frozen. Source tests and a downloadable archive still do not
prove installed multi-session behavior.

## Current checkpoint

Bind qualification only after the Axios v2 evidence and baseline changes are reviewed, integrated, and the resulting exact dev candidate is frozen. This issue owns supported-package and installed multi-session evidence across macOS, Linux, and Windows. Marketplace publication and post-publish verification remain solely in #1189.

The release coordinator requires full qualification on Metal and Windows Vulkan and marks standalone Linux Vulkan qualification and the checksum-bound quality adjunct nonblocking. #1221 retains its stricter closure bar: the same F/archive must also pass the standalone Linux Vulkan full qualification and the existing three-repeat quality contract. Those receipts do not create a broader public performance or answer-quality claim.

## Dependencies

- The #1220 portion of the same PR supplies the accepted server protocol,
  source cutover, observable identity fields, and frozen timing/threshold
  record before qualification begins on the unchanged candidate.
- The accepted Windows source-build/staging repair from PR #1262 and #1230's exact-run identity must be represented truthfully before their evidence can enter release closeout.
- Accepted output from this issue feeds #1233 and final qualification #1189.

## Required proof

On each supported OS, under one OS account, install the exact package and
launch two fresh, independently started plugin sessions against different
repositories. Exercise the ordinary plugin path in both:

- race a cold connect/spawn and prove one endpoint listener and one server
  instance; use separate process and engine telemetry to prove one engine
  owner/native worker, one load generation, and one model load. Endpoint or
  server count alone is not model-load proof;
- run query and bulk work from different repositories and prove bounded queue
  state, query preference between bulk batches, FIFO within each class, bulk
  resumption under the preregistered mixed workload, no project-, repository-,
  tenant-, or scope-based round-robin, useful retry conditions, and no project
  leakage. Do not claim a bounded-starvation guarantee under arbitrary
  sustained query traffic;
- kill one client and prove its queued work and leases are reclaimed without
  disrupting the other;
- crash the server and prove the next client elects one replacement;
- inject a model-worker no-progress stall and prove the independent watchdog
[...]

PR #1382, opened 2026-07-22
Wait for Windows package server idle exit

TheGreenCedar:
## Context

Exact Windows-only run `29894398392` proved the package and all isolated MCP calls, then failed when `TemporaryDirectory` tried to remove the unpacked candidate while its per-user server was still inside the frozen 60-second idle window. The Actions runner later reported that exact `codestory-cli.exe` as an orphan.

## What changed

- pin an identity-safe native wait handle as soon as the initial exact packaged server is authenticated;
- retain the final qualification authority after every package and qualification scenario;
- at the final temporary-directory boundary, launch the exact packaged CLI directly as an observational MCP client and require a final live server snapshot on successful proof paths;
- verify and pin a replacement final server if qualification changed its PID/start identity;
- close the final client, then drain every owned exact-process waiter through the normal 60-second idle path with the existing 2.5-second qualification grace;
- require exit code `0` from the final authenticated server; a nonzero earlier exit is accepted only when a different final PID/start identity proves that process was superseded, and is retained as `superseded_process_exit` rather than clean idle;
- run the guard after runtime or qualification failure as well as success, close every owned handle even when observation, host close, or wait raises, and preserve the primary proof failure with cleanup failures attached as secondary context;
- retain successful cleanup evidence in the package result;
- cover clean exit, bounded timeout, abnormal termination rejection, explicit superseded-exit classification, physical Windows executable unlock/removal, and primary-error preservation in the checker self-test;
- record the release-proof behavior in `CHANGELOG.md`.

There is no process-name enumeration, broad kill, forced cleanup, or cleanup relaxation.

## Review

Focus on `ExactProcessExitWaiter`, `FailurePreservingTemporaryDirectory`, `pin_temporary_package_server`, `wait_for_final_temporary_package_server`, and the guarded runtime/qualification block in `main`. The initial and final handles remain bound to their kernel processes even if a PID is reused; all handles are drained and closed at the outer package boundary. Only a different authenticated final identity can supersede a historical fault exit.

## Verification

- `python .github/scripts/check-packaged-agent-proof.py --self-test` — passed on macOS at exact head
- the same self-test — passed on physical Windows 11 Pro x64 at exact head; copied/running `proof-process.exe` exited through the waiter, forced termination was rejected, the same nonzero exit was accepted only under explicit superseded semantics, the synthetic cleanup failure preserved its primary proof error, and post-check counts were process `0`, directory `0`
- Python compile check — passed on macOS and physical Windows at exact head
- `git diff --check` — passed on macOS and physical Windows at exact head
- workflow policy — passed; workflows are unchanged from the earlier 276/276 mutation proof
- focused PR checks are running for this exact head
- full current-candidate Windows package replay remains the live acceptance boundary

Exact head: `a1f3e23ebef44cd699bf37fdb7f9c4b0ec4a09a8`
Tree: `47a26318599e317c3cb169a5d7c9ecb1ed0b2448`
Base: `f56e1a65d0ed59a4f0e864078b7c9ce7b98ff0ee`

## Risk and non-claims

The successful Windows hosted-package path now takes up to the frozen idle timeout plus 2.5 seconds before archive cleanup. The retained old PC archive predates the native-manifest ordering and qualification-path fixes, so it cannot authenticate this cleanup fix. A fresh physical-Windows build was attempted: the shared `sccache` service panicked during compiler-version IPC, and bypassing it under the SSH session reached Windows `os error 448` (`untrusted mount point`) before compilation. No fresh package was produced. Source checks and the focused physical self-test do not prove a full current-candidate cleanup replay, protected Vulkan, candidate-installed behavior, Mac/Metal, answer quality, promotion, publication, or release readiness.

Closes #1381
Refs #1372
Refs #1221
Refs #1189

### DIFF
--- a/.github/scripts/check-packaged-agent-proof.py
+++ b/.github/scripts/check-packaged-agent-proof.py
@@ -9487,6 +9487,37 @@ def self_test() -> None:
         and exit_evidence["process_start_id"] == exit_process_start,
         "exact process normal-exit wait self-test failed",
     )
+    if os.name == "nt":
+        with tempfile.TemporaryDirectory(
+            prefix="codestory-executable-cleanup-self-test-"
+        ) as cleanup_raw:
+            cleanup_root = Path(cleanup_raw)
+            cleanup_executable = cleanup_root / "proof-process.exe"
+            shutil.copy2(
+                Path(os.environ["SystemRoot"]) / "System32" / "ping.exe",
+                cleanup_executable,
+            )
+            cleanup_process = subprocess.Popen(
+                [str(cleanup_executable), "-n", "2", "127.0.0.1"],
+                stdin=subprocess.DEVNULL,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+            )
+            cleanup_start = process_start_identity(cleanup_process.pid)
+            cleanup_waiter = ExactProcessExitWaiter(
+                cleanup_process.pid,
+                cleanup_start,
+                self_target_os,
+            )
+            try:
+                cleanup_waiter.wait(5_000)
+            finally:
+                cleanup_waiter.close()
+                cleanup_process.wait(timeout=5)
+        require(
+            not cleanup_root.exists(),
+            "exact process exit wait left its Windows executable locked",
+        )
     timeout_waiter = ExactProcessExitWaiter(
         self_pid,
         self_start_id,
--- a/.github/scripts/check-packaged-agent-proof.py
+++ b/.github/scripts/check-packaged-agent-proof.py
@@ -3269,7 +3269,7 @@ class ExactProcessExitWaiter:
             self.close()
             raise
 
-    def wait(self, timeout_ms: int) -> dict:
+    def wait(self, timeout_ms: int, *, require_clean_exit: bool = True) -> dict:
         require(timeout_ms > 0, "exact process exit wait requires a positive timeout")
         if self.target_os == "windows":
             kernel = ctypes.windll.kernel32
@@ -3294,10 +3294,11 @@ class ExactProcessExitWaiter:
                 bool(kernel.GetExitCodeProcess(self.handle, ctypes.byref(exit_code))),
                 f"could not read exact process {self.pid} exit code",
             )
-            require(
-                exit_code.value == 0,
-                f"exact process {self.pid} exited abnormally with code {exit_code.value}",
-            )
+            if require_clean_exit:
+                require(
+                    exit_code.value == 0,
+                    f"exact process {self.pid} exited abnormally with code {exit_code.value}",
+                )
         else:
             exit_code = None
             deadline = time.monotonic() + (timeout_ms / 1000)
@@ -3323,13 +3324,18 @@ class ExactProcessExitWaiter:
                 time.sleep(0.01)
         return {
             "status": (
-                "normal_idle_exit"
+                (
+                    "normal_idle_exit"
+                    if exit_code.value == 0
+                    else "superseded_process_exit"
+                )
                 if self.target_os == "windows"
                 else "observed_exit"
             ),
             "pid": self.pid,
             "process_start_id": self.expected_start_id,
             "exit_code": exit_code.value if exit_code is not None else None,
+            "clean_exit_required": require_clean_exit,
             "timeout_ms": timeout_ms,
         }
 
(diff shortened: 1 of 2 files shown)


TheGreenCedar (commented):
Review of exact head `315eb95794c8a84c9d5e49694eaa3aff40388205` against `dev/codestory-next` at `f56e1a65d0ed59a4f0e864078b7c9ce7b98ff0ee`. The PR moved to `0d61bf0` during review; this is deliberately not an approval of that newer head.

[P1] Move the unpacked-server wait after follow-up qualification

The new wait runs inside `prove_runtime` (around lines 4368-4388), but `main` can then call `produce_qualification_evidence` (lines 11425-11435) using the same unpacked `qualification_cli`. The required scenario order ends with `worker_stall`; that scenario observes and uses a replacement server, then returns without waiting for it to idle out. On a full protected/calibration Windows package proof, this later replacement is another process executing the temporary package binary, so leaving the outer `TemporaryDirectory` can still hit the same `WinError 5` after the earlier waiter has already completed. The final exact server created by all follow-up qualification work needs to be pinned and allowed to exit before archive cleanup.

[P1] Do not label every signaled process handle a normal idle exit

`ExactProcessExitWaiter.wait` treats `WAIT_OBJECT_0` as sufficient and unconditionally records `status: normal_idle_exit`. A crash or external termination also signals the process handle, so the checker can attest normal idle shutdown when the server actually died abnormally during the window. Read the exit code from the same pinned handle after the wait and fail closed unless it is the expected successful idle-shutdown status.

TheGreenCedar (commented):
Review of exact head `cc503651bd3ebc2bc08a44b07ed2456007c98f62` against `dev/codestory-next` at `f56e1a65d0ed59a4f0e864078b7c9ce7b98ff0ee`.

The two prior happy-path findings are closed: final observation now follows `produce_qualification_evidence`, uses the final qualification namespace, pins the verified native process handle before closing the observational client, and checks exit code `0` from that same handle before recording `normal_idle_exit`.

[P1] Do not accept pre-observation absence as clean cleanup proof

At lines 3525-3552, a missing `embedding_server` snapshot returns `status: already_absent` and the checker continues. That path has no authenticated final PID/start identity, no pinned process handle, and no exit code, so a server that crashes or exits nonzero after qualification but before the observational read is accepted even though the PR and changelog say abnormal termination fails closed and clean idle exit is required. Every affected Windows non-installed lane has just used or respawned the package server, so absence at this boundary should either fail closed or be backed by an earlier pinned handle that can still prove the exact process exit status.

[P1] Run final cleanup when runtime or qualification proof raises

The call at lines 11585-11591 is reached only after both `prove_runtime` and `produce_qualification_evidence` return successfully. If an MCP assertion or qualification scenario fails after starting the unpacked server, the waiter is skipped, the outer `TemporaryDirectory` can reproduce `WinError 5`, and that cleanup error can mask the actual proof failure. This is especially concrete because `produce_qualification_evidence` publishes `_server_cleanup_control` before running the scenarios. The final exact-process wait needs exception-safe placement at the temporary-directory boundary, while preserving the original failure; the helper should also close an already-created waiter if `host.close()` itself raises.

TheGreenCedar (commented):
Exact-head review of `7f4b0f6b9ace787867281ba40cf719bbfd553715` against `dev/codestory-next` at `f56e1a65d0ed59a4f0e864078b7c9ce7b98ff0ee`.

One blocking finding:

[P1] Do not require the qualification fail-stop owner to exit normally - `.github/scripts/check-packaged-agent-proof.py:3656`

On Windows hosted/protected qualification runs, `prove_runtime` pins the initial package server at line 4265 before `produce_qualification_evidence` runs. That qualification necessarily includes `server_crash` (and its default publication-fault producer also issues `crash_server` at line 5073); the native transport implements that action with `std::process::abort()` in `crates/codestory-cli/src/embedding_server_transport.rs:557`. The final cleanup then calls `ExactProcessExitWaiter.wait()` for every pinned entry, and that method rejects any nonzero exit code. The intentionally aborted original owner therefore makes the otherwise successful exact Windows qualification fail even when the final replacement is authenticated and exits cleanly. Keep waiting and closing every pinned handle, but apply the clean-exit assertion only to the final authenticated server (or explicitly classify expected qualification fail-stop identities before waiting). The focused self-test passes, but it does not exercise this integrated qualification path.

The other three prior gates are closed in source: cleanup now runs after qualification/replacements, success requires an authenticated final snapshot, and exception/host-close/wait paths drain and close the retained waiters while preserving a primary proof error.

TheGreenCedar (commented):
Fourth exact-head review of `a1f3e23ebef44cd699bf37fdb7f9c4b0ec4a09a8` against `dev/codestory-next` at `f56e1a65d0ed59a4f0e864078b7c9ce7b98ff0ee`.

No findings. Approved at this exact head.

The prior P1 is closed. On a successful Windows package proof, the final snapshot is authenticated through `server_snapshot`, its exact PID/start identity and executable are pinned, and only that final identity is waited with `require_clean_exit=True`. A prior pinned identity can accept a nonzero exit only when it differs from that authenticated final identity; the waiter classifies that nonzero result as `superseded_process_exit`. The returned cleanup evidence is independently required to name the final server's `normal_idle_exit`, exit code `0`, and `clean_exit_required=True`.

There is no abnormal-exit relaxation that can produce successful evidence. The broader exception-cleanup path runs only after a captured runtime or qualification failure, returns no cleanup evidence, drains and closes every retained waiter, and then re-raises the original failure with any observation, wait, host-close, or temporary-directory cleanup problem attached as secondary context. Final waiting remains after all qualification and replacement work, and successful proof cannot use an unauthenticated absent-server path.

Verification at this head: checker self-test passed, Python compilation passed, `git diff --check` passed, and the focused GitHub checks completed successfully. The fresh full Windows package replay remains the live acceptance boundary already stated by the PR.

Issue #1577, opened 2026-07-28
Dead-client workers each hash the whole executable before applying pressure


TheGreenCedar:
## Problem

`crates/codestory-cli/src/embedding_qualification/worker/operations/dead_client.rs:62` calls
`NativeEmbeddingClientTransport::capture()` **inside** each spawned thread. `capture()`
(`crates/codestory-cli/src/embedding_server_transport.rs:81`) resolves `current_exe()` and runs `sha256_reader`
over the whole executable.

`spawn_dead_client_request` is called once per `parameters.query_count` and once per `parameters.bulk_count`
(`:31-42`), so the CLI binary is hashed once per worker thread. Two consequences:

1. **Wasted work that scales with worker count** — every thread recomputes an identity that is identical for all of
   them, since they all run from the same executable.
2. **It contaminates what the worker measures.** The comment at `:59-61` says these workers deliberately avoid
   product deadlines so as not to "add cancellation retries to the pressure this worker is intended to measure".
   But each thread now spends a full executable hash *before* it issues its request, so admitted-request pressure
   arrives late and staggered rather than concurrently. The shape of the pressure is an artifact of the hashing, not
   of the load being modelled.

## Outcome

Capture the transport identity **once**, before the workers are spawned, and share that one captured identity across
every worker thread.

## Acceptance

- The executable is hashed once per dead-client operation regardless of `query_count` + `bulk_count`.
- Every worker thread uses the same captured identity; nothing weakens the identity check itself, and a capture
  failure still prevents the worker from issuing requests rather than proceeding unauthenticated.
- Workers still apply pressure concurrently — the fix must not serialise them behind a shared lock held across the
  request.
- A test pins the capture count: N workers must produce exactly one capture.

## Boundaries

Do not change what `capture()` verifies. Do not introduce a process-wide cache in
`embedding_server_transport.rs` — the reuse belongs to this operation, whose threads share one executable by
construction. Related to the de-hot-pathing of the same 150MB hash done in WP4.

Not release-blocking.

PR #1578, opened 2026-07-28
share one dead-client executable capture across worker threads

TheGreenCedar:
## Context

`run_dead_client_load` spawns one thread per `query_count` and one per `bulk_count`, and each thread called
`NativeEmbeddingClientTransport::capture()` itself. `capture()` resolves `current_exe()` and sha256-hashes the whole
executable, so the binary was hashed once per worker — twice, in fact, because the thread body captured a transport
for its clock and then `run_raw_protocol_exchange_with_input` captured a second one for the exchange.

The redundancy is the smaller problem. These workers deliberately skip product deadlines so cancellation retries do
not pollute the pressure they exist to measure, but a full executable hash in front of every request made that
pressure arrive late and staggered instead of concurrently — the shape of the load became an artifact of the hashing.

## What changed

- `dead_client.rs` captures once, before the first spawn, and shares that single capture with every worker through an
  `Arc`. The fan-out now lives in `spawn_dead_client_workers`, which takes the capture and the per-worker request as
  parameters so a test can pin the capture count without asserting on timing.
- `protocol.rs` gains `run_raw_protocol_exchange_with_transport`, the exchange against an identity the caller already
  captured. `run_raw_protocol_exchange_with_input` is now that function plus its own capture, so no existing caller
  changes behaviour.
- A stale sentence in the bench runner's `dead_client_setup_timeout` doc comment ("fans out one captured transport per
  held request") now describes the shared capture. The budget itself is unchanged — it is an upper bound, and this
  change only shortens the ramp it bounds.

Nothing about what `capture()` verifies changed, and no process-wide cache was added to
`embedding_server_transport.rs`; the reuse belongs to this operation, whose threads share one executable by
construction.

## How to review

Start at `spawn_dead_client_workers`. The capture runs before the loop and `?` propagates, so a capture failure fails
the operation instead of leaving threads that return early and silently skip their request — the old thread body's
`Err(_) => return`. Workers still run free: they share an `Arc`, not a lock, and nothing is held across the request.

## Verification

Both directions on the new tests:

- `cargo test -p codestory-cli --lib --locked dead_client` — 2 passed; 0 failed.
- Reverted only the source shape (capture moved back inside each thread, early return on error) with the tests
  untouched: both failed — `every_worker_shares_one_capture_and_applies_pressure_concurrently` reported
  `left: 5, right: 1` captures for 5 workers, and `a_failed_capture_fails_the_operation_without_issuing_a_request`
  reported five spawned `JoinHandle`s where the operation should have failed. Restored: 2 passed; 0 failed.

Gates:

- `cargo test -p codestory-cli --locked --test architecture_contracts` — 14 passed; 0 failed.
- `cargo test --workspace --locked` — counts in the verification comment on this PR.
- `cargo clippy -p codestory-cli --all-targets --locked` — clean.

## Risk

Low and confined to the qualification worker. The concurrency assertion is a rendezvous, not a duration: a worker
leaves only once every worker has entered its request, so a change that serialises the fan-out fails instead of merely
running slower. Its 30s wait exists solely so such a regression fails rather than hangs; a concurrent fan-out never
approaches it.

## Follow-up

None. `run_raw_protocol_exchange_with_input` still captures per call for its single-exchange callers, which is correct
for them.

Closes #1577

### DIFF
--- a/crates/codestory-bench/src/bin/codestory_embedding_qualification/scenarios/artifact/runner/process.rs
+++ b/crates/codestory-bench/src/bin/codestory_embedding_qualification/scenarios/artifact/runner/process.rs
@@ -422,8 +422,8 @@ pub(super) fn load_establishment_timeout(phase: &str) -> Result<Duration> {
 /// `client_death_lease_active` wait must see the lease plus the admitted and
 /// held query/bulk work in one snapshot, and that bounds a queue-seeding
 /// phase, not a single snapshot: the freshly spawned dead client first pays
-/// its contract connect and spawn-convergence allowances, then fans out one
-/// captured transport per held request before the seeded depths become
+/// its contract connect and spawn-convergence allowances, then fans every held
+/// request out over one captured transport before the seeded depths become
 /// visible, while every poll of the wait is itself a fresh observe worker
 /// spending part of the snapshot allowance. This is the same phase shape as
 /// mixed_queue's gated seeding, which already bounds a strictly larger 64+64
--- a/crates/codestory-cli/src/embedding_qualification/worker/operations/dead_client.rs
+++ b/crates/codestory-cli/src/embedding_qualification/worker/operations/dead_client.rs
@@ -1,10 +1,11 @@
-use super::super::protocol::run_raw_protocol_exchange_with_input;
+use super::super::protocol::run_raw_protocol_exchange_with_transport;
 use super::ANTI_IDLE_PROTOCOL_DEADLINE_MS;
 use anyhow::{Result, bail};
 use codestory_retrieval::{
     AwakeMonotonicClock, EmbeddingClientTransport, EmbeddingQualificationParameters,
     PerUserEmbeddingClient, SidecarRuntimeConfig,
 };
+use std::sync::Arc;
 use std::time::Duration;
 
 const CLIENT_DEATH_LEASE_HOLD_MS: u64 = 600_000;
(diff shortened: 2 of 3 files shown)


TheGreenCedar:
## Verification

Exact head: `5434e0494956181b9dc4a38aedbd118640fe7521`, based on `origin/dev/codestory-next` at
`7df8467eb8cb794f4d90d3457b1574c4e03e447f`. Host: Apple Silicon macOS (Darwin 25.5.0).

### Mandatory gates

| Command | Result |
| --- | --- |
| `cargo test -p codestory-cli --locked --test architecture_contracts` | 14 passed; 0 failed; 0 ignored |
| `cargo test --workspace --locked` | 58 test binaries; 2333 passed; 0 failed; 29 ignored; exit 0 |
| `cargo clippy -p codestory-cli -p codestory-bench --all-targets --locked -- -D warnings` | clean |

The workspace run above is on this exact head. An earlier `cargo test --workspace --locked` on the same source, before
the bench doc-comment edit, also exited 0.

### Both directions on the new tests

Forward, `cargo test -p codestory-cli --lib --locked dead_client`:

```
test ...dead_client::tests::a_failed_capture_fails_the_operation_without_issuing_a_request ... ok
test ...dead_client::tests::every_worker_shares_one_capture_and_applies_pressure_concurrently ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 264 filtered out
```

Reverted: only the source shape went back to the pre-fix behaviour — the capture moved inside each spawned thread with
the old `Err(_) => return` early exit — with both tests left exactly as written. Same command:

```
test ...a_failed_capture_fails_the_operation_without_issuing_a_request ... FAILED
test ...every_worker_shares_one_capture_and_applies_pressure_concurrently ... FAILED

---- every_worker_shares_one_capture_and_applies_pressure_concurrently ----
assertion `left == right` failed: every dead-client worker must share one executable capture
  left: 5
 right: 1

---- a_failed_capture_fails_the_operation_without_issuing_a_request ----
a failed capture must fail the operation: [JoinHandle { .. }, JoinHandle { .. }, JoinHandle { .. },
JoinHandle { .. }, JoinHandle { .. }]

test result: FAILED. 0 passed; 2 failed
```

Five workers produced five captures instead of one, and a failing capture returned five spawned workers instead of
failing the operation — the two defects this PR fixes. Restoring the source returned the run to 2 passed; 0 failed.

### Not claimed

No qualification scenario was executed on hardware. The dead-client worker's real behaviour under the
`client_death_lease_active` phase is unchanged in contract and is covered by the existing bench lane, which this PR
does not run.

Issue #1575, opened 2026-07-28
Make packet sufficiency prove each requirement, not each role


TheGreenCedar:
## Outcome

Replace the hand-written claim taxonomy with per-requirement evidence predicates so a resolved exact path needs its own proof-bearing claim.

## Provenance

Work exists on the unpushed local branch `codex/wp9c-sufficiency-truthfulness-r2`, produced by an earlier multi-agent wave and never built or
tested by anyone until now. It was assessed and then independently judged; the diff is real, but it does not merge
as-is. This issue tracks bringing it to a mergeable state.

## What must be true before this merges

vidence predicates, which means re-deriving the mapping from scratch. The genuinely good parts are separable and should be salvaged as their own commits.

---

## Prioritised list — what must be true before this could merge

1. **Requirement matching must be per-requirement, not per-role.** Give each `FlowRequirement` an evidence predicate (allowed `PacketEvidenceRole` set, or a citation predicate) instead of `roles.contains(&requirement.role)`. Non-negotiable: probes A, D, E, F must all return to `Partial`.
2. **Restore `request_interceptor_management` as a cited-role requirement.** Re-wire `packet_citation_owns_interceptor_management` into sufficiency. It is currently dead code in that path and it is what the axios gate turns on.
3. **Restore `format_errors` / `FlowRole::ErrorOrFallback`** by extending the evidence-role mapping to reach it, not by deleting the requirement. If nothing can carry it, that is a retrieval gap to fix, not a requirement to drop.
4. **Reinstate the 7 falsified refusal tests** — the three `*_remain_distinct`, both `form_validation_*_is_partial`, `sql_looking_claim_text_without_structural_citations_stays_partial`, `runtime_formatting_output_claims_do_not_cover_error_fallback_role` — rewritten with *cited* claims so they exercise the new mechanism rather than the retired one.
5. **Add a same-role distinctness invariant test**: for any flow whose table contains two requirements sharing a `FlowRole`, evidence closing one must leave the other missing. This is the single property whose absence let the regression through.
6. **Add regression tests pinned to the three holdout prompts** in `benchmarks/tasks/holdout-retrieval/`, asserting `Partial` when the named component is uncited. These are the lane's own acceptance criteria and belong in the unit suite.
[...]

Issue #1572, opened 2026-07-28
Readiness now reports a healthy fresh manifest-published store as stale


TheGreenCedar:
## Problem

#1557 (merged as PR #1562) made the readiness projection consult storage-derived staleness so it could stop over-claiming. It now **under-claims**: a healthy, fresh, manifest-published store is classified `stale_publication`, so `doctor` reports

> semantic stale: the published retrieval index is stale or degraded for the current contract. Run `codestory-cli retrieval index --refresh full` before trusting packet/search evidence.

for an index that is fine — and the advised refresh republishes the identical manifest, so the warning never clears. That is the exact failure mode the test's own comment says it was written to prevent.

## Reproduction

```
cargo test -p codestory-cli --lib doctor_semantic_check_is_healthy_for_fresh_manifest_published_store
```

- passes at `a7420167` (before #1562 merged)
- fails at `aabf04b6` (dev, after): `left: "warn"  right: "ok"`

`crates/codestory-cli/src/app/tests/lifecycle/packet_diagnostics.rs:158`.

## Cause

In `crates/codestory-runtime/src/search_publication.rs`, `stale_publication` was manifest-shape only; it now also fires on `storage_admission_refusal_reason_for_runtime`. A store that has a published manifest but whose symbol-doc / dense-anchor aggregates are zero is read as stale rather than as consistent-with-its-manifest.

## Why it was not caught

#1562's verification ran `cargo test -p codestory-runtime` (858 pass). The regressed consumer lives in `codestory-cli`, which was never run. The runtime crate is not the blast radius of a change to a projection that CLI surfaces consume.

## Acceptance

- The doctor semantic check reports `ok` for a healthy fresh manifest-published store.
- #1557's original fix is preserved: readiness must still refuse to claim `semantic_ready` for a publication sidecar admission would refuse after a core-only refresh or an interrupted incremental run. Do not simply revert.
- Regressions cover both directions — over-claiming and under-claiming — and live where a future change will actually run them.
- `cargo test --workspace` is green, not just the runtime crate.

Refs #1557

TheGreenCedar:
**Correction to this issue's premise.**

The investigation showed the projection was right and the *fixture* was wrong. `retrieval_state_from_storage_for_runtime` is byte-identical between the #1562 merge and the fix, and the true refusal was:

```
retrieval_manifest_stale: sidecar_symbol_doc_count_changed: manifest=8 current=0
```

[...]

Issue #1574, opened 2026-07-28
Rank grounding from repository-derived signal, not benchmark heuristics


TheGreenCedar:
## Outcome

Rank compact grounding by architecture evidence derived from the repository and report typed orientation uncertainty, preserving deterministic bounded budgets.

## Provenance

Work exists on the unpushed local branch `codex/wp9b-ranking-rebuild-r2`, produced by an earlier multi-agent wave and never built or
tested by anyone until now. It was assessed and then independently judged; the diff is real, but it does not merge
as-is. This issue tracks bringing it to a mergeable state.

## What must be true before this merges

ect deleted. The replacement-budget invariant has no direct equivalent; the new `diversify_root_order` takes no limit at all, and the general property `smaller_limit_results_are_an_exact_prefix_of_larger_limit_results_for_one_candidate_set` is stronger. Acceptable.
- `public_surface_question_keeps_short_pascal_case_named_anchor` survives with holdout literals scrubbed (`getLatestSocialEntries`/`getElsewhereFeed` → `getQuellRecords`/`getQuellStream`) — same assertion.

New tests are substantive property tests over synthetic identifiers (`zqAlpha`, `quell`, `zarq`, `vorbex`), plus 11 in `root_rank.rs`. `entry_evidence` is a real classifier, not a stub. One soft spot: the token-closure invariant exempts roles `original_question` and `named_anchor`, so an expansion reintroduced under either role name evades it.

## Would merging break anything currently passing?

Yes, exactly one: `scripts/tests/lint-retrieval-generalization.test.mjs` → `tasks about other repositories still ban their symbols`. Green on dev, red after merge. Plus the non-conflicting compile break, which git will not warn about.

## Prerequisites before this can merge, in priority order

1. **Replace the `baseName === "tests.rs"` exclusion with a `#[cfg(test)] mod X;`-aware rule.** Keying on filename is an unreviewable bypass; my `ranking_probe` reproduction must fail the lint.
2. **Add a guard test for that hole** — plant a non-`cfg(test)` `tests.rs` under `crates/codestory-runtime/src` carrying a banned name and a corpus path, assert the lint rejects it. The existing `linter_scans_every_runtime_source_file_for_holdout_names` does not cover it.
[...]

Issue #1573, opened 2026-07-28
Derive the retrieval generalization ban from the corpus without losing coverage


TheGreenCedar:
## Outcome

Delete benchmark-shaped retrieval steering and derive the lint's banned corpus from benchmarks/tasks/** instead of a hand list.

## Provenance

Work exists on the unpushed local branch `codex/wp9a-terms-lint-r2`, produced by an earlier multi-agent wave and never built or
tested by anyone until now. It was assessed and then independently judged; the diff is real, but it does not merge
as-is. This issue tracks bringing it to a mergeable state.

## What must be true before this merges

int exit 0, lint node tests 2/2. The one CLI failure is pre-existing on dev, reproduced there. What merging breaks is not a test — it is the lint's coverage.

## What must be true before this merges (prioritised)

1. **Restore the 13 lost bans.** Either derive bare mid-path/prose-claim tokens, or re-add an explicitly-labelled residual literal list for what derivation cannot reach. A merge that ships this as-is means `HiArgs`, `chinook`, `server.c`, and `--with-holdout-clone` can be written into `crates/codestory-retrieval/src/` with CI green.
2. **Add a regression test that pins the coverage floor.** A test that plants dev's full pre-branch literal set and asserts every one is reported. Without it, item 1 can silently regress again.
3. **Fix `linter_catches_current_holdout_literals_in_production`** to assert on all 9 of its planted literals, not 3. It currently provides false assurance.
4. **Ratchet the pending file.** Either a CI check that the inventory's total never grows relative to the merge base, or a required `reason` + issue link per entry. As it stands it is a self-service excuse file with no automated bound and no test.
5. **Wire `scripts/tests/lint-retrieval-generalization.test.mjs` into `retrieval-engine-smoke.yml`,** or port the two self-repository tests into the guard suite. The rule the merge resolution chose is otherwise untested in CI.
6. **Justify the `repo_text.rs` inversion with evidence.** Two source files that the Sourcetrail architecture question used to surface no longer do. Either show a benchmark run demonstrating the net effect is neutral-or-better, or state plainly in the PR that retrieval quality on that task regressed and why that is the accepted price.
7. **Broaden `linter_bans_holdout_repository_names_on_identifier_boundaries`** to iterate every repo name in `benchmarks/tasks/**`. It would immediately surface the `express`/`http`/`requests` exclusion, which someone should then rule in or out deliberately.

[...]

PR #1581, opened 2026-07-28
close each flow requirement with its own cited evidence

TheGreenCedar:
Closes #1575

## What changed

Packet sufficiency decided requirement coverage by the requirement's `FlowRole`. A claim whose *wording* produced that role closed every requirement wearing it, so two requirements in the same flow — a client's request finalization and its transport send, a command loop's event loop and its network input — were closed together by one piece of evidence, and prose alone could close either.

Each `FlowRequirement` now carries an **evidence predicate**: either an allowed set of `PacketEvidenceRole` values, or a structural check over the cited anchor (`EvidencePredicate::CitedRoles` / `CitedCarrier`). A requirement is covered only by a proof-bearing claim that cites evidence matching *that* requirement. `StructuralLanguagePolicy` and the claim-text→`FlowRole` derivation are gone; `FlowRole` remains as a diagnostic label.

Specifically:

- `request_interceptor_management` is back in the sufficiency path, wired through `packet_citation_owns_interceptor_management`. This is what the axios holdout gate turns on.
- `format_errors` / `FlowRole::ErrorOrFallback` are kept, with an evidence carrier of their own rather than being deleted.
- The new `packet_evidence_carriers` module holds the per-requirement structural checks. Every carrier reads only the citation.

Alongside it, sufficiency stops publishing what it cannot prove:

- `covered_claims` and `avoid_opening_paths` come from proven claims only. A claim the same packet lists as ineligible is no longer republished as verified, and the file behind it is no longer advertised as not worth opening.
- An uncited sentence is ineligible ("claim carries no cited evidence").
- The SQL diagnostic-evidence bypass is removed: a synthetic source scan is a text match and no longer promotes a verdict.
- Every task class — not only `ArchitectureExplanation` — holds each resolved exact path to its own proof-bearing claim, with a per-path gap, a bounded gap budget (`MAX_EXACT_PATH_CLAIM_GAPS`), and the path-specific follow-up placed at the front of the follow-up list so the command cap cannot drop it.

One term-detection fix was required: `packet_terms_indicate_shell_install_dispatch_flow` accepted "command"/"function" as a shell signal, so a command-server prompt raised shell requirements that no citation in such a repository could ever close. It now needs an actual shell term.

## Invariants

Three tests that the previous shape could not express:

- `the_requirement_inventory_matches_the_requirement_tables` — a checked-in inventory of every requirement. The old invariant asked whether *any* evidence role could carry a requirement's `FlowRole`, which a failing lane could satisfy by deleting the requirement; this one fails on removal too, so the only way past it is an edit a reviewer reads.
- `every_requirement_has_evidence_that_can_close_it` — each requirement has a witness anchor that closes it, so an unclosable requirement is caught instead of reporting partial forever.
- `requirements_sharing_a_flow_role_stay_separable_by_evidence` — for every pair of requirements sharing a `FlowRole` inside one flow, each one's witness must leave the other missing. This is the property whose absence let one citation close a whole role.

Plus `holdout_prompts_stay_partial_when_the_named_component_is_uncited`, pinned to the three prompts in `benchmarks/tasks/holdout-retrieval/`, with the opposite direction for axios in `holdout_axios_interceptor_evidence_closes_the_interceptor_requirement`.

## Tests changed rather than added

Every test that proved a requirement through claim wording had to be rewritten with cited evidence; the claim text and the prompt are unchanged in each case, only the fixture now carries the anchor the claim rests on. Two changed direction, for stated reasons:

- `sql_synthetic_source_scan_table_and_foreign_key_cover_schema_requirements` → `sql_synthetic_source_scan_evidence_never_covers_schema_requirements`. Same prompt, same three source-scan anchors; the removed bypass is exactly what made them count as proof. The scan is still reported with its concrete role as evidence worth following up.
- `packet_sufficiency_uses_selected_plan_role_probes` → `planned_role_probes_stay_sufficiency_gaps_until_evidence_covers_their_requirement`. Same four HTML/JS anchors; none is evidence for a native constraint, a custom validator, or a submit guard, so all three requirements stay open and every planned probe for them stays a gap. It now also asserts the three requirements are missing and not covered.

`compact_truncated_packet_retains_proof_provenance_counts` keeps the same seven-tier fixture and now derives its expected `covered_claims` count from which anchors are sufficiency-eligible, instead of asserting all seven.

## Both directions

Each behavioural change was reverted in place and the corresponding test observed to fail:

| Mutation | Failing test |
| --- | --- |
| `request_interceptor_management` given its `Dispatch` sibling's predicate | `requirements_sharing_a_flow_role_stay_separable_by_evidence`; the axios holdout goes **Sufficient** with the interceptor uncited |
| `format_errors` deleted from the table | `the_requirement_inventory_matches_the_requirement_tables` |
| `format_errors` given `CitedRoles(&[])` | `every_requirement_has_evidence_that_can_close_it` |
| exact-path check restricted to `ArchitectureExplanation` | `every_task_class_needs_a_proof_bearing_claim_for_each_resolved_exact_path` |
| uncited-claim branch removed | `a_claim_without_cited_evidence_cannot_satisfy_sufficiency` |
| `covered_claims: supported_claims` | `a_claim_the_packet_reports_as_unproven_is_never_published_as_covered` |
| single combined exact-path gap restored | `more_uncovered_exact_paths_than_the_gap_budget_are_summarized_not_dropped`, `retained_route_tracing_packet_reports_the_unproven_route_instead_of_sufficient` |
| old shell-install term set restored | `a_command_server_prompt_is_not_a_shell_install_prompt` |

## Gates

- `cargo test -p codestory-cli --test architecture_contracts` — 14 passed.
- `cargo test --workspace --no-fail-fast` — green except `doctor_semantic_check_is_healthy_for_fresh_manifest_published_store`, which fails identically at plain `origin/dev/codestory-next` (issue #1572, fix in flight elsewhere). Not touched here.
- `cargo clippy --workspace --all-targets` — clean.

## Not done

`runtime_retrieval_quality.py` was not run. It is a **verifier** for an evidence artifact, not a producer: it needs a packaged release build, network clones of axios/redis/ripgrep at their pinned refs, and three repeats across the release quality modes. That artifact comes from `packaged-platform-proof.yml` (`--retrieval-quality-evidence`), so the corpus evidence has to be attached from a CI run before merge. The holdout acceptance criteria themselves are now pinned in the unit suite so they cannot regress silently between corpus runs.


TheGreenCedar:
**Holding this rather than landing it.** Recording why, and what would change the answer.

Round 3 is the largest genuine narrowing of the three rounds — adversarial acceptances fell 73% in the generated name space and 83% in the path space, six requirements are fully cleaned, and the two-factor framing is the right design. That is real progress and none of it is wasted.

But the verifier reached **NOT READY** on a specific class: a false-safe `Sufficient` is still reachable in five of six flows using ordinary symbols. The defect this commit's own title names — the path deciding which step a citation proves — survives in three carriers via the directory and in every `CitedRoles` requirement via the file name. The commit body and CHANGELOG currently assert the opposite of what the code does.

**Why that is disqualifying here when a residual gap was acceptable on the sibling lanes.** #1573 and #1574 are a lint and a ranking: a gap there means a missed token or slightly worse ordering, and both still catch strictly more than dev. This lane decides whether CodeStory reports `Sufficient` to the user. A leak means it claims evidence it does not have — which is the exact failure #1200 and #1351 exist to remove. Shipping a narrower version of that is still shipping it.

**The three smallest changes identified as closing what was found:**
1. `role_survives_without_its_directory` — strip the file *name* as well, keeping only the extension, applying the same monotone narrowing to the whole path. Then give `request_terminal`'s `BufferedIo`/`EventOutputProcessing` roles a name-side route the way `IndexingWorkQueue` got `index`+`entrypoint`, or they become unreachable.
2. `is_form_validation_surface` and `belongs_to_logging` — use `names_token`, not `names_or_path_token`, matching what `citation_owns_log_record_creation` already does and what the module header already claims.
3. `names_site_build_object` — drop `file`/`files`/`document`/`documents`, which are what `readFile`/`writeFile`/`document.write` ride in on, and revisit the newly added `run`/`start`.

Also to fix before it lands: the commit body and CHANGELOG claims that do not match the code.

Not closing the PR — the work is close and the direction is right.

TheGreenCedar:
HOLD at exact head `f5dafa66887e009f7f162897901e6419d45ee7db`.

Two full packet-verdict reproductions still return `Sufficient` with an unproved requirement:

- Real `Logger.addRecord` evidence plus unrelated `DefaultHandler.process` from `src/http/default_handler.rs` closes `handler_processing`.
- Real `Buffer` evidence plus unrelated `DatabaseSource.read` and `TelemetrySink.write` closes `buffered_read_write`.

In both cases the packet has no gap and publishes the unrelated citation paths as already covered. The generated off-subject sweep misses this because it tests residual evidence by itself; pairing a declared residual with valid evidence for its sibling closes the whole flow.

The PR is back in draft. An independent verifier is reproducing these exact shapes before remediation.

TheGreenCedar:
Independent confirmation on exact head `f5dafa66887e009f7f162897901e6419d45ee7db`:

`cargo test --locked -p codestory-runtime independent_residual_ -- --nocapture` ran two temporary full-verdict regressions; both failed because the actual status was `Sufficient` rather than the required `Partial`.

- `Logger.addRecord` + unrelated `DefaultHandler.process`: empty gaps; `handler_processing` falsely covered.
- `Buffer` + unrelated `DatabaseSource.read` / `TelemetrySink.write`: empty gaps; `buffered_read_write` falsely covered.

The invariant checks residual evidence in isolation. It does not compose a valid sibling citation with the residual citation, so it misses the exact way the whole flow becomes falsely safe. The draft hold remains.

TheGreenCedar:
## Independent adversarial repair — exact head ready for counterexample review

Head: `739037370d64106976eb03529198747c120f2137`
Tree: `c32edb3e07c24c6936ae7347b744e938e9e2a6a4`
Base of this repair: `f5dafa66887e009f7f162897901e6419d45ee7db`

The previous head was executed before editing and returned false-safe `Sufficient` verdicts for both residual shapes the round-five report called safe:

- real `Logger.addRecord` plus unrelated `DefaultHandler.process`;
- real `Buffer` plus unrelated `DatabaseSource.read` and `TelemetrySink.write`.

Both regressions now return `Partial` and name the missing step (`handler_processing` or `buffered_read_write`). A full-verdict matrix does the same for all six carrier flows, including object-word + step-verb anchors filed under `.vue`, `.svelte`, and HTML/markup surfaces.

### Hostile mutations executed

Each mutation was applied in place, the named test was executed, and the accepted implementation was then restored:

| Reintroduced defect | Result |
| --- | --- |
| generic `source` / `sink` / `stream` may prove buffered reads or writes | buffer verdict regression failed because the packet became `Sufficient` |
| structural adjectives such as `default` / `processing` may establish logging ownership | logger verdict regression failed because the packet became `Sufficient` |
| a `forms/` directory may establish form ownership for script/component anchors | six-flow verdict test failed on the form case with `Sufficient` |
| a `site/` directory may establish static-site ownership | six-flow verdict test failed on the static-site case with `Sufficient` |
| a `mapping/` directory may establish object-mapper ownership | six-flow verdict test failed on the mapper case with `Sufficient` |
| an `fmt/` directory may establish formatter-argument ownership | six-flow verdict test failed on the runtime-formatting case with `Sufficient` |

### Focused verification on the pushed source

- carrier-flow generated sweep: 3 passed, including the six full-verdict cases and every plausible owned surface
- evidence-carrier unit suite: 6 passed
- flow-requirement suite: 19 passed
- packet-sufficiency suite: 102 passed
- architecture contracts: 14 passed
- focused runtime clippy with warnings denied: clean
- runtime formatting check and `git diff --check`: clean

No workspace proof, packaged proof, calibration, hardware proof, or merge was started. This draft remains an active parallel lane and is awaiting counterexample-only review of the exact head above.

TheGreenCedar:
## Adversarial repair round 2 — exact head ready for re-verification

Head: `97fb528c63255022034b4d3e4c888237e98b39dd`
Tree: `2a157ee654ea5ce792202c26dd610733d94b1b7e`
Rejected head: `739037370d64106976eb03529198747c120f2137`

The independent verifier rejected the previous head on two executed failures, both reproduced locally before this edit:

1. Near-prefix geometry symbols carrying argument/error step words returned false-safe `Sufficient` with no missing requirements.
2. The retrieval generalization lint rejected benchmark-shaped production commentary in the new carriers.

### Repair

Runtime-format ownership now uses explicit whole-token vocabularies. The carrier no longer treats every token beginning with `format` as an inflection. A generated regression crosses five neighboring suffix families with six argument-step words, six failure-step words, and Rust/Vue/Svelte/HTML surfaces: 240 unrelated anchors must close neither requirement. The full six-flow verdict test now includes several of those neighboring families and requires both `format_arguments` and `format_errors` to remain named gaps.

The production commentary was rewritten without protected corpus/example identifiers. The existing lint catches this directly; no lint exemption or pattern weakening was added.

### Red/green evidence

Before the repair:

- `cargo test -p codestory-runtime --lib runtime_formatting_rejects_words_that_only_neighbor_the_format_stem -- --nocapture` failed on the first generated argument anchor.
- `cargo test -p codestory-runtime --lib every_carrier_flow_is_partial_and_names_the_unproved_step -- --nocapture` failed because the near-prefix packet was `Sufficient`, with empty gaps and missing sets.
- `node scripts/lint-retrieval-generalization.mjs` failed on five protected production matches.

On the pushed tree:

- evidence-carrier suite: 7 passed
- flow-requirement suite: 19 passed, including the generated cross-surface carrier sweep
- packet-sufficiency suite: 102 passed; all six carrier-flow packets are `Partial` and name their uncovered step(s)
- architecture contracts: 14 passed
- retrieval generalization lint: clean
- retrieval generalization Node contracts: 4 passed
- focused runtime clippy with warnings denied: clean
- runtime formatting and diff checks: clean

No workspace proof, package proof, calibration, hardware dispatch, merge, or release action was started. The PR remains draft pending counterexample-only verification of this exact head.

TheGreenCedar:
Independent exact-head verification accepted `97fb528c63255022034b4d3e4c888237e98b39dd` (tree `2a157ee654ea5ce792202c26dd610733d94b1b7e`).

- clean and pushed (`+0/-0`);
- all six carrier flows remained `Partial`;
- `FormationArgs` / `FormationError` and an independent 30-case Rust/Vue/Svelte/HTML/XHTML near-prefix matrix found no false-safe verdict;
- the generated off-subject sweep passed at least 5,000,000 comparisons;
- `handler_processing` remained `Partial` with that exact gap;
- `buffered_read_write` remained `Partial` with that exact gap;
- production generalization lint passed across 289 production files, 222 protected non-Rust files, and 1,714 patterns.

No counterexample remains from this verification pass. The PR stays sequenced independently from the frozen 0.16.3 candidate and does not gate that release.

PR #1582, opened 2026-07-28
[codex] derive the generalization lint's banned corpus and delete benchmark-shaped retrieval steering

TheGreenCedar:
Closes #1573

## Context

Search-plan term extraction carried tables written from the benchmark corpora: a
domain-noun list, a role-spec table mapping invented role names to holdout
vocabulary, and inference rules that read a question for `data`/`accessed`/
`application` and added `storage`/`persistence`. The generalization lint that was
supposed to catch that kind of thing kept its own banned corpus as an
88-entry hand list, so it could only ever ban what somebody had already noticed.

This deletes the steering and makes the lint read its corpus out of
`benchmarks/tasks/**` and the benchmark harness surfaces on every run.

## What changed

**Steering deleted** (`crates/codestory-runtime/src/search_terms.rs`,
`search_plan.rs`, `crates/codestory-retrieval/src/lexical_index.rs`)

- `SEARCH_PLAN_SYMBOL_TERMS` and `SEARCH_PLAN_ROLE_SPECS` are gone, with the
  planner paths that consumed them.
- `add_search_plan_inferred_architecture_terms`,
  `drop_search_plan_brand_terms_for_content_flow`, and the exec/payload flow
  predicates are gone. Search-plan terms now come from the asked question only.
- `search_plan_broad_explanation_prompt_with_architecture_terms` became
  `search_plan_prose_flow_prompt`, which asks whether the question names an
  identifier instead of counting matches against an architecture-noun table.
- `SEARCH_PLAN_STOPWORDS` stays; it is language-level filler and encodes no
  repository.
- The lexical command-token path no longer special-cases `codex`.

**Lint corpus derived** (`scripts/lint-retrieval-generalization.mjs`)

Repository names, task ids, prompts, claims, expected symbols, expected files,
qualified names, fixture names, and harness repositories are read out of the
checked-in corpora on every run: 1,706 patterns, up from 1,420. Adding a task
manifest extends the ban without editing the lint. Three derivations were added
in this round on top of the earlier work:

- **Prose identifiers.** `HiArgs`, `aeProcessEvents`, and `search_parallel` are
  written only into claim sentences, never into `expected_symbols`. Identifier
  words are now taken from claim prose, restricted to lower-first camelCase and
  snake_case — the two casings that only come from code. PascalCase is not safe
  to take: `SQLite` is this product's own storage engine, `JavaScript` is a
  language, `URLSession` and `ValidityState` are platform APIs.
- **Corpus file names.** `ae.c` is too short for the path-window rule, so corpus
  basenames become boundary-carrying identity markers, minus build-layout stems
  and minus stems built from words the product writes itself.
- **Member symbols.** `app.use` is a whole symbol below the eight-character
  descriptive floor; `receiver.member` gets its own floor because two joined
  identifiers cannot be a word at any length.

**Self-subject rule tightened.** The merge kept the derived rule
(`benchmarkRepoIsProduct`) over dev's explicit slug set. As written it matched
any crate-name token, so a holdout repository called `store`, `runtime`, or
`bench` would have claimed the self-subject exemption and switched the lint off
for itself. It now matches this repository's own name, derived as the prefix
every crate shares.

## Prerequisites from the previous review

**1. The 13 lost bans are restored.** Five came back by derivation
(`aeProcessEvents`, `search_parallel`, `ae.c`, `app.use`, and `app.use`'s split
form). Eight are in `residualBannedLiterals`, an explicitly labelled list where
each entry states why derivation cannot reach it — repositories retired from the
corpus (`freelancer`, `traderotate`, `ElsewhereFeed`, `--with-holdout-clone`), a
nickname whose derivation would also ban `animate` and `commons` (`chinook`), an
owner segment (`mdn`), a PascalCase prose word (`HiArgs`), and a file name whose
stem the product writes itself (`server.c`). 88 hand-written literals became 8
reasoned ones.

**2. The coverage floor is pinned.**
`linter_still_reports_every_ban_it_had_before_the_corpus_was_derived` plants all
93 of dev's pre-branch literals (90 line-scan, 3 split for the compact scan) and
requires a report for every one.

**3. `linter_catches_current_holdout_literals_in_production` asserts on all 9**
of its planted literals, not 3.

**4. The pending inventory is bounded and attributable.** Every surface now
carries a `reason` (at least 60 characters) and the issue tracking its deletion,
and the declared `total_markers` must equal the number of markers listed — no
slack in either direction, so growth is a reviewable diff that raises a stated
number. The lint exits 2 on any violation.

**5. `scripts/tests/lint-retrieval-generalization.test.mjs` runs in
`retrieval-engine-smoke.yml`,** and gained two tests for the self-subject rule
the merge chose (a holdout named after one of our crates must still be banned;
this repository's own name must still be exempt).

**6. The `repo_text.rs` inversion is a stated regression, not a claimed
improvement.** `architecture_repo_text_window_preserves_non_crate_source_surfaces`
used to assert that `StorageAccess.h` and `StorageAccessProxy.cpp` surfaced for
the Sourcetrail architecture question. They reached that window because
`add_search_plan_inferred_architecture_terms` read
("data" + "accessed" + "application") and added the words `access`, `storage`,
and `persistence`. Both files are expected files of a holdout task, and the rule
that fetched them was written from that task. **No benchmark run is claimed
here.** Running the holdout suite needs materialised clones and an agent loop
this lane cannot produce, so this PR does not assert the net effect is
neutral-or-better. It asserts the mechanism is gone and accepts that this
question's recall drops with it: a rule that can only pay on the corpus it was
written against is not retrieval quality, it is a memorised answer. The test now
asserts the two files stay out and says so in a comment.

**7. `linter_bans_holdout_repository_names_on_identifier_boundaries` iterates
every repository named under `benchmarks/tasks/**`** — 47 names — and requires
each to be banned unless it appears in `CORPUS_NAMES_RULED_OUT_OF_THE_BAN` with a
reason. The test also fails on a stale ruling, so an exclusion cannot outlive its
justification. The rulings:

| Name | Ruling |
| --- | --- |
| `CodeStory`, `codestory` | this repository is its own benchmark subject |
| `express` | `codestory-indexer/src/framework_routes.rs` extracts Express routes as a parser-backed product feature and must name the framework it parses |
| `fmt` | `std::fmt`; banning it forbids every `Display`/`Debug` impl |
| `http` | the product speaks HTTP in its own adapters |
| `requests` | plural of a word the product's request plumbing is built from |

## Pending inventory: 192 to 204

The widened derivation reached 12 benchmark-family surfaces that were previously
invisible, all in files the inventory already covers: `path.ends_with("ctx.py")`,
`path.ends_with("logger.php")`, `path.ends_with("mapper.cs")`, `typemap.cs`,
`scaffold.py`, `io_client.dart` in `packet_scoring.rs`; `div#app`,
`searchparallel`, `processinputbuffer` in `packet_claim_profiles.rs`; `divapp` in
`packet_scoring.rs` and `packet_sufficiency.rs`. These are real steering, newly
visible rather than newly created. Deleting them changes packet ranking and
belongs with the other 52 entries in `packet_scoring.rs`, so they are recorded
under the new reason+issue discipline instead. **This is growth in the inventory,
chosen deliberately**: the alternative was to keep derivation narrow enough not
to see them, which is the behaviour the inventory exists to stop.

## How to review

1. `scripts/lint-retrieval-generalization.mjs`: `residualBannedLiterals` (what
   derivation cannot reach, and why), `proseIdentifierMarkers` /
   `codeCasedProseWord`, `corpusFileNameMarkers`, `memberSymbolMarker`,
   `productRepositoryNames`, `loadPendingSurfaces`.
2. `crates/codestory-runtime/tests/retrieval_generalization_guard.rs`:
   `PRE_DERIVATION_BAN_FLOOR` and `CORPUS_NAMES_RULED_OUT_OF_THE_BAN`.
3. `crates/codestory-runtime/src/search_terms.rs` for what was deleted.

## Verification

Head `4d3c7534`, base `origin/dev/codestory-next` at `f5c2ec54`.

These were re-run against the pushed head rather than carried over from an
earlier one; the counts below moved as the identifier-word-break pass landed.

| Command | Result |
| --- | --- |
| `node scripts/lint-retrieval-generalization.mjs` | exit 0 — 1,706 patterns, 204 pending markers in 14 files |
| `node --test scripts/tests/lint-retrieval-generalization.test.mjs` | 4/4 pass |
| `cargo test -p codestory-cli --test architecture_contracts` | 14 passed, 0 failed |
| `cargo test --workspace --no-fail-fast` | 2,349 passed, 0 failed |
| `cargo test -p codestory-runtime --test retrieval_generalization_guard` | 32 passed, 0 failed |
| `cargo clippy --locked -p codestory-runtime -p codestory-retrieval --all-targets` | no warnings |
| `node .github/scripts/check-workflow-policy.mjs` | passed |
| `node .github/scripts/check-doc-links.mjs` | passed (75 files, 266 links) |
| `git diff --check` | clean |

### Both-directions proofs

Each new gate was run with its change reverted and observed to fail.

| Gate | Reverted | Observed failure |
| --- | --- | --- |
| Coverage floor | dropped `residualBannedLiterals` from `bannedPatterns` | `deriving the corpus lost these bans: ["freelancer", "traderotate", "chinook", "mdn", "ElsewhereFeed", "HiArgs", "server.c", "--with-holdout-clone"]` |
| Coverage floor | disabled prose identifiers, file names, member-symbol floor | `lost these bans: ["aeProcessEvents", "search_parallel", "ae.c", "app.use", "\"app\", \".use\""]` |
| Item 3 (all 9 literals) | dropped `residualBannedLiterals` | `lint failure should report current holdout literal HiArgs` — the previous 3-literal assertion still passed in that state, which is the false assurance it was flagged for |
| Item 7 (all repo names) | removed the `express` ruling | `these corpus repository names are not banned and are not ruled out: ["express"]` |
| Pending: reason required | removed one `reason` | exit 2, `reason must be at least 60 characters saying why the surface still exists` |
| Pending: issue required | replaced one `issue` with prose | exit 2, `issue must be a CodeStory issue URL tracking the surface's deletion` |
| Pending: total bound | added a marker without raising the cap | exit 2, `declares total_markers 204 but lists 205` |

While writing the coverage-floor test I found and fixed a false negative in it:
fixture names `floor_1.rs` and `split_floor_1.rs` nest as substrings, so a report
for the split fixture read as coverage for the plain one and hid two lost bans.
The two families are now `floor-N.rs` and `joined-N.rs`.

No existing test was weakened or deleted. Two were widened
(`linter_catches_current_holdout_literals_in_production` 3 to 9 assertions,
`linter_bans_holdout_repository_names_on_identifier_boundaries` 1 name to 47);
both still cover their original inputs.

### Pre-existing failure, resolved upstream

`doctor_semantic_check_is_healthy_for_fresh_manifest_published_store` (#1572)
failed on the original base `dcd1f31f` and on this branch while it was based
there. It passes at plain `origin/dev/codestory-next` `7df8467e`, which contains
the fix (`69d05b74`). This branch was merged up to `7df8467e`; the workspace run
above is on the merged head and is fully green.

## Risk

- Lint coverage widened to 1,706 patterns. A false positive shows up as a CI
  failure naming the pattern and line, not as a silent behaviour change.
- Deleting the search-plan tables changes which subqueries a broad architecture
  question produces. `crates/codestory-runtime/src/tests/search_plan.rs` was
  rewritten with the deletions; `repo_text.rs` records the one known recall loss
  above.
- The pending inventory grew by 12. Every entry names its reason and issue, and
  the declared total now bounds it.

## Follow-up

The 204 inventoried surfaces in agent packet code and architecture coverage
scoring are the remaining benchmark-shaped retrieval steering. They are recorded,
reported on every run, and bounded, but not deleted.

## Provenance

Nothing on this lane is merged anywhere. `codex/wp9a-terms-lint-r2` (`e52b99e4`)
is an ancestor of **this branch**, not of `dev/codestory-next` —
`git merge-base --is-ancestor e52b99e4 origin/dev/codestory-next` exits 1. Every
commit between `origin/dev/codestory-next` and the head is new to this PR and in
scope for review; none of it has been reviewed anywhere else.

An earlier revision of this description implied that work had been "brought
across from a merged branch". That was wrong and is corrected here.


TheGreenCedar:
**This survives the wp9a rewrite.** Verified by execution against `origin/codex/wp9a-generalization-lint` (PR #1582), not by reading.

The rewrite replaces `taskSubjectIsThisRepository` with `benchmarkRepoIsProduct` and tightens *which* names count as product (from any crate-name token down to the repository's own derived name). But it keeps `repo?.name` as an input:

```js
function benchmarkRepoIsProduct(repo) {
  return [repo?.name, ...repoUrlSlugs(repo?.url)]
    .filter((value) => typeof value === "string")
    .some((value) => productRepositoryNames.has(value.split("/").pop().toLowerCase()));
}
```

Replicating its own derivation (`productRepositoryNames` = `['codestory']`, the common prefix of every `crates/` entry) and running the predicate:

```
impostor (name=codestory, url=https://github.com/axios/axios) exempted? true
real CodeStory task                                           exempted? true
foreign axios task                                            exempted? false
```

So a holdout task declaring `repo.name: "codestory"` while pointing anywhere still contributes zero banned markers, exactly as on dev. The tightening addressed a *different* hole (a holdout named `store`/`runtime`/`bench` claiming the exemption) and left this one open.

**Fix:** drop `repo?.name` from the inputs and match only `repoUrlSlugs(repo?.url)`. Confirmed this keeps the real case working, because `https://github.com/TheGreenCedar/CodeStory` yields the slug `codestory` from the URL itself — the name field is not load-bearing for the legitimate case, only for the impostor one.

**Status on dev:** the hole is live but unexploited — all 34 tasks under `benchmarks/tasks/**` were checked and none declares `name: codestory` with a foreign URL. So this is a latent hole to close in PR #1582 rather than an active exemption.

Issue #1591, opened 2026-07-29
Serialize retrieval generalization fixture mutations across test processes


TheGreenCedar:
Parent: #1179

## Context

The CodeStory 0.16.3 proof-only release run 30411881352 passed workflow policy, release preflight, source resolution, all three packaged builds, and the macOS Metal, Windows Vulkan, and Linux Vulkan hardware proofs. Its full source gate failed only in `retrieval_generalization_lint_script_exits_clean_with_extra_fixture_root`.

The guard suite uses a process-local Rust mutex around lint subprocesses. The hosted full-source gate runs tests with cargo-nextest, which executes individual tests in separate processes. At the same time, `linter_scans_a_tests_dot_rs_that_no_parent_declares_cfg_test` temporarily writes `crates/codestory-runtime/src/ranking_probe/tests.rs` into the real checkout to prove the default scan catches an unlisted shipped module. The clean-scan process observed that planted file and rejected its deliberate `benchmarks/tasks` marker.

## Outcome

Make every test that reads or temporarily mutates the default lint roots coordinate across processes, so parallel test execution cannot turn one hostile fixture into another test result.

## Scope

- Replace process-only serialization with a cross-process-safe test guard shared by every default-root lint invocation.
- Add a regression that exercises the process boundary that failed in Actions.
- Preserve the current default scan coverage, including orphan `tests.rs` modules and every runtime source path.

## Acceptance criteria

- [ ] A hostile source fixture planted by one test process cannot be observed by a concurrent clean default-root scan.
- [ ] The regression fails when serialization is reduced back to a process-local mutex.
- [ ] `retrieval_generalization_lint_script_exits_clean_with_extra_fixture_root` passes under cargo-nextest parallel execution.
- [ ] The focused guard suite and the required workspace and architecture-contract gates pass.
- [ ] The lint gains no new exclusion, allowance, or fail-open behavior.

## Verification

- `cargo nextest run -p codestory-runtime --test retrieval_generalization_guard`
- `cargo test -p codestory-runtime --test retrieval_generalization_guard`
- `cargo test -p codestory-cli --test architecture_contracts`
- `cargo test --workspace --no-fail-fast`

## Non-goals

- Do not weaken benchmark-corpus detection.
- Do not exempt `ranking_probe/tests.rs` or any other production-shaped path.
- Do not change release workflow claims or hardware gates.

TheGreenCedar:
Coordinator lane started.

[...]

PR #1592, opened 2026-07-29
[codex] serialize retrieval lint fixtures across nextest processes

TheGreenCedar:
## Context

Closes #1591
Refs #1179

CodeStory blocks benchmark-shaped retrieval steering with a lint that scans the real production roots. Two guard tests temporarily plant hostile Rust files in those roots to prove an unlisted runtime file and an orphan `tests.rs` module cannot escape the scan.

The guard serialized lint subprocesses with a Rust `Mutex`. That works under the standard test harness, but the 0.16.3 source proof uses cargo-nextest, which runs each test in its own process. Release run 30411881352 failed when a clean default-root scan observed `crates/codestory-runtime/src/ranking_probe/tests.rs` while another process had planted it.

## Outcome

Default-root lint reads and temporary source plants now share one checkout-scoped advisory file lock in addition to the existing thread lock. Parallel cargo-nextest processes cannot observe each other's fixture state. If a planter is killed before its destructor runs, the next lock holder removes only a known fixture path carrying the exact test-ownership marker before scanning.

## What changed

- Hold the existing thread guard and an `fs4` exclusive lock at `target/retrieval-generalization-guard.lock` around every lint subprocess.
- Keep the lock across the complete baseline, plant, hostile scan, and normal cleanup sequence.
- Mark the two real-tree fixtures as test-owned and recover only those exact paths after abnormal process exit.
- Add real process regressions for lock exclusion, crash recovery, and the non-deletion boundary.
- Keep poison recovery for the thread mutex so one assertion still reports as one failure.

No production source, lint rule, workflow, or release claim changed.

## How to review

1. `lint_script_lock_for`: confirm recovery occurs only after acquiring the checkout-scoped process lock.
2. `recover_stale_planted_sources`: confirm both the exact known path and exact ownership marker are required.
3. `run_default_lint_with_planted_source`: confirm the guard covers baseline through fixture cleanup.
4. The process regressions: remove the exclusive lock or recovery call, and try widening cleanup beyond either ownership proof.

## Verification

Head `ef028057d16768e230a3f169cff19e34bda532e7`, base `2292cff92144935e3592027ab41195e3644b0de0`.

- Pre-fix hostile reproduction at the base: staggered cargo-nextest processes made the planter pass in 22.643s and the clean scan fail in 11.399s on the exact planted `createApplication` / `benchmarks/tasks` payload.
- Lock mutation: removing `FileExt::lock_exclusive` makes `lint_script_lock_serializes_independent_test_processes` fail because a second process acquires the lock early.
- Crash mutation: removing the recovery call makes `lint_script_lock_recovers_a_fixture_after_process_exit` fail with the planted source still present.
- Real SIGKILL reproduction: after the actual planter created `ranking_probe/tests.rs`, killing its Rust process returned 137; the next clean process removed the marker-owned file and passed.
- Independent exact-head refutation repeated that process and confirmed no orphan process, file, or directory remained.
- Independent cleanup mutations proved the boundaries are live: adding an unknown path or removing the marker gate makes the preservation regression fail.
- `cargo nextest run --locked -p codestory-runtime --test retrieval_generalization_guard --no-fail-fast` — 46 passed, 2 skipped; independently repeated with the same result.
- `cargo test --workspace --locked` — every workspace test binary and doctest passed, including the full 46-test standard guard.
- `cargo test --locked -p codestory-cli --test architecture_contracts` — 14 passed.
- Focused clippy with `-D warnings`, per-crate formatting, and `git diff --check` — clean.
- Remote saga guard, draft source checks, rustdoc, and retrieval contracts — green.

## Risk or follow-up

The lock and recovery are test-only. Recovery is intentionally narrower than a general stale-file cleanup: it ignores every unknown path and every known path without the exact ownership marker. The empty lock file may remain under ignored `target/` output.

### DIFF
--- a/crates/codestory-runtime/tests/retrieval_generalization_guard.rs
+++ b/crates/codestory-runtime/tests/retrieval_generalization_guard.rs
@@ -1,24 +1,68 @@
 //! Ensures the retrieval generalization lint script stays runnable from the workspace root.
 
-use std::fs;
+use fs4::fs_std::FileExt;
+use std::fs::{self, File, OpenOptions};
 use std::path::{Path, PathBuf};
 use std::process::{Command, Output};
 use std::sync::{Mutex, MutexGuard, OnceLock};
+use std::thread;
+use std::time::{Duration, Instant};
 use tempfile::TempDir;
 
 static LINT_SCRIPT_LOCK: OnceLock<Mutex<()>> = OnceLock::new();
+const LINT_SCRIPT_LOCK_HELPER_PATH: &str = "CODESTORY_LINT_SCRIPT_LOCK_HELPER_PATH";
+const LINT_SCRIPT_LOCK_HELPER_READY: &str = "CODESTORY_LINT_SCRIPT_LOCK_HELPER_READY";
+const LINT_SCRIPT_LOCK_HELPER_RELEASE: &str = "CODESTORY_LINT_SCRIPT_LOCK_HELPER_RELEASE";
 
-/// Serialises the lint subprocesses. The lock guards no shared state -- only the
-/// cost of running many `node` processes at once -- so a test that panics while
-/// holding it has corrupted nothing. Recovering from the poison rather than
-/// propagating it keeps one real failure reported as one failure: without this,
-/// the first genuine assertion turns every later test into a `PoisonError`
-/// panic, and the failure list says nothing about how much is actually broken.
-fn lint_script_lock() -> MutexGuard<'static, ()> {
-    LINT_SCRIPT_LOCK
+struct LintScriptGuard {
+    _thread_guard: MutexGuard<'static, ()>,
+    process_lock: File,
+}
+
+impl Drop for LintScriptGuard {
+    fn drop(&mut self) {
+        let _ = FileExt::unlock(&self.process_lock);
+    }
+}
+
+/// Serialises lint subprocesses in this test binary and across independent test
+/// processes. Cargo-nextest gives each test its own process, so the Rust mutex
+/// alone cannot protect a clean default-root scan from another test temporarily
+/// planting a hostile source file in the checkout.
+///
+/// Neither lock protects product state. Recovering a poisoned thread lock is
+/// therefore safe and keeps one real assertion failure from turning every later
+/// guard test into a `PoisonError` panic.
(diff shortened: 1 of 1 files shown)


TheGreenCedar:
Independent exact-head review disposition: HOLD.

The normal cross-process race is fixed at `e717787068c34ebb166a28b1276c95a61d5fe673`, but the hostile process-death check found a residual in the same shared-checkout contract:

- the verifier started the real orphan-`tests.rs` planter and waited for `crates/codestory-runtime/src/ranking_probe/tests.rs`;
- killing that Rust test process released the advisory lock, but `PlantedSource::drop` never ran;
- the 107-byte hostile fixture remained in the production tree;
- a separate clean scan acquired the released lock and failed on the stale `createApplication` / `benchmarks/tasks` payload.

The verifier removed only that test-created residue and confirmed the exact-head worktree was clean. This draft remains held while the same lane makes planted fixtures crash-recoverable through an explicit ownership marker and adds a process-exit regression. No merge or release proof will start from the current head.

TheGreenCedar:
Crash-recovery follow-up is pushed at exact head `ef028057d16768e230a3f169cff19e34bda532e7`.

The actual planter was started, observed writing `crates/codestory-runtime/src/ranking_probe/tests.rs`, and then SIGKILLed. The next clean process removed the exact marker-owned residue and passed. Recovery now requires both a known fixture path and the exact ownership marker; focused tests also pin that an unmarked known path and a marked unknown path survive.

The PR remains draft. The verifier who found the kill gap is re-running the hostile shape on this exact head; the full guard, architecture contracts, and workspace-wide gate follow only if that independent pass accepts it.

TheGreenCedar:
Exact head `ef028057d16768e230a3f169cff19e34bda532e7` is accepted.

- Independent cancellation review: PASS, including the real killed planter and three cleanup-boundary mutations.
- Full cargo-nextest guard: 46 passed, 2 skipped; independently repeated.
- Full workspace gate: every test binary and doctest passed, including the 46-test standard guard.
- Architecture contracts: 14 passed.
- Remote saga, source, rustdoc, and retrieval checks: green.

No fixture file, directory, lint process, or test process remains. This is now mergeable into `dev/codestory-next`.

Issue #1735, opened 2026-08-02
[v0.17 CI-LANES] Run merged proof suites in per-PR CI lanes


TheGreenCedar:
## Context

Several proof suites added by merged program PRs run in no per-PR CI lane — they execute only in the dispatch-gated workspace suite:

- EV-FIX (#1718): `packet_evidence::tests`, the packet-sufficiency provenance test, the packet-batch default test, the `search_scoring` lane assertions, and the CLI `stdio_catalog` schema test
- READY-A (#1727): lease tests in `codestory-runtime` `services::activation_tests`, `codestory-cli` `stdio_transport` unit tests
- HOOK (#1672): the 35 `repository_hooks` boundary tests in `codestory-workspace`
- W1.SEC (#1643): the workspace reader's own unit tests

Per-PR coverage today is rust-ci's six pinned invocations (`rust-ci.yml:117-123`), `linux-contracts` in `retrieval-engine-smoke.yml:154-175`, and `crate-durability.yml:80-82` (store + indexer only). A regression in any of the suites above would merge green and surface only at the next workspace-suite dispatch.

## Contract

- Extend the per-PR lanes (the `linux-contracts` job is the cheap insertion point) so each suite above runs on every PR touching its crate.
- Keep the repo's workflow rules: rust-ci cargo invocations are SHA-256'd into cache keys (mirror any change atomically), and workflow changes are mirrored into `release-claims.json` `workflow_policy` in the same PR.

## Done when

- A deliberate break in each listed suite fails PR CI (spot-verified for at least one test per suite).
- Cache keys and workflow policy remain consistent.

Refs #1718. Refs #1727. Refs #1672. Refs #1643. Refs #1179.

Issue #1727, opened 2026-08-02
[v0.17 READY-A] Reuse a valid ready lease across unchanged broad calls


TheGreenCedar:
Source package **READY-A** in the v0.17.0 consolidated remediation program.

## Contract

Reuse a runtime-owned ready lease for repeated broad calls on the same unchanged
logical project. The fast path must return before constructing a second runtime
context or starting retrieval finalization/native embedding. Admission uses the
runtime's bounded configuration plus complete core/retrieval publication and
producer identities; missing, corrupt, or changed pointer evidence replaces the
lease exactly once.

Preserve resident controller state when project, storage, and complete core
publication identity are unchanged. Replacement remains fail-closed old-or-new:
no caller may observe mixed core/retrieval state, and concurrent drift callers
join one replacement activation.

## Parent proof tracker

#1599 retains the exact packaged Linux CPU replay with a concurrent tiny second
project. This child owns the remaining source implementation and deterministic,
execution-shaped tests; its PR closes only this child.

## Dependencies and order

W1.SEC / #1643 is merged. This source layer precedes EV-5 / #1701 at the shared
`stdio_transport.rs` hotspot and becomes the fan-out foundation for RUN-B / #1660
and RUN-C / #1706.

## Program boundaries

- Use a `codex/*` branch and target `dev/codestory-next`.
- Centralize immutable runtime configuration identity below the CLI so the lease
  and adapter cannot disagree.
- Probe a cloned lease outside the coordinator lock, then relock and revalidate
  target, revision, and lease identity before reuse.
- Capture the verified source-freshness state and gap that produced the lease,
  but do not turn reuse admission into strict readiness or an O(repository)
  fingerprint pass. Arbitrary post-ready filesystem edits have no cheap live
  signal before EV-7b's observer; READY-A must not claim otherwise. Existing
  packet/readiness freshness remains fail-closed, and an authoritative observer
  epoch may extend this lease later.
- Keep source, package, protected-host, installed-runtime, and publication proof
  distinct.
- Preserve atomic publication, pinned readers, crash recovery, freshness,
  fail-closed evidence, observational status/doctor, accelerated embedding,
  proof-owned cleanup, and serial Cargo execution.
- Do not touch product version or release surfaces.

## Done when

- A second same-project broad call constructs no runtime context, preserves the
  activation operation/revision/attempt and core/retrieval/server/model identity,
  and invokes neither finalization nor native embedding.
[...]

Issue #1718, opened 2026-08-02
[v0.17 EV-FIX] Restore evidence tier honesty and packet timeout contracts


TheGreenCedar:
Package **EV-FIX** in the v0.17.0 consolidated remediation program.

## Why this exists

The post-merge audit of EV-A / PR #1710 and EV-3 / PR #1713 found deterministic acceptance regressions that must land before EV-5:

- rank export injects lexical score into pure dense candidates, so DenseSemantic is unreachable and dense-only evidence can become sufficiency-eligible;
- typed file/range conversion can erase intrinsic StructuralText or ExactSource evidence;
- remaining production `build_search_hit` paths do not bind explicit lane provenance before evidence classification;
- `same_file_name_affinity` is absent from public packet provenance;
- the packet batch default was reduced from 18,000 ms to 1,500 ms while the orchestrator still uses 18,000 ms, and explicit ask budgets remain capped at 1,500 ms.

## Contract

Repair the merged evidence foundation as one contiguous corrective layer:

1. Export zero lexical/semantic/graph score for lanes that did not produce that provenance. A pure dense candidate classifies DenseSemantic and is ineligible for sufficiency; affinity is never graph evidence.
2. Preserve node-intrinsic StructuralText and ExactSource stamps when attaching typed FILE/range target data. Do not upgrade arbitrary FILE nodes.
3. Bind explicit provenance and classify exactly once on every production search-hit path that can feed packet evidence; origin fallback cannot silently manufacture ResolvedGraph eligibility.
4. Preserve `same_file_name_affinity` as a public non-graph provenance label.
5. Keep the packet-level default at 18,000 ms. Keep 1,500 ms only as the unspecified single-sidecar default. Explicit `latency_budget_ms` values honor the documented 1,000–120,000 ms contract without an invented 1,500 ms cap.
6. Re-run the EV-A production-order matrix and packet/search evaluation with a per-category cause table. No unexplained partial-to-sufficient movement.

## Dependencies

EV-A / #1644 and EV-3 / #1648 are already merged. This corrective PR is the new bottom of the evidence stack and blocks EV-5 / #1701 and EV-4 / #1245.

## Program boundaries

- The implementation PR uses a `codex/*` branch, targets `dev/codestory-next`, and carries `Closes #N`.
- Keep source, package, protected-host, installed-runtime, and publication proof distinct.
[...]

Issue #1672, opened 2026-08-02
[v0.17 HOOK] Install hooks only through the effective trusted hooks path


TheGreenCedar:
Package **HOOK** in the v0.17.0 consolidated remediation program.

## Contract

Deliver W8.9 with one Rust-owned effective-hooks resolver and mutation seam:

- resolve `core.hooksPath` with bounded system, global/XDG, environment,
  repository-local, and worktree precedence without spawning Git;
- retain W1.SEC's no-follow capture for repository-controlled metadata, disable
  repository-driven includes, and fail `hooks_config_unresolved` whenever
  precedence cannot be proven;
- resolve the default to the validated common repository's `hooks` directory and
  relative configured paths against the selected worktree root;
- accept only destinations contained by the W1-validated project, git, or common
  roots; external paths, `/dev/null`, traversal, ambiguous includes, symlink or
  reparse components, and missing proof are typed non-installable outcomes;
- preflight all three hook targets before mutation and splice only verified
  regular executable shell dispatchers through a pinned no-follow directory.

The JavaScript hook manager remains the user-facing wrapper and dirty-marker
entrypoint, but it must call the Rust resolver/mutator through an explicit
installed/dev CLI seam. It must not duplicate Git config interpretation or
download a CLI during status.

## Dependencies and stack shape

W1.SEC / #1643 and SRC-B / #1656 are merged. HOOK branches from current dev and
remains independent of READY-A/EV-5; stacking it above either would create a
false dependency.

## Program boundaries

- The implementation PR uses a `codex/*` branch, targets
  `dev/codestory-next`, and carries `Closes #1672`.
- Keep source, package, protected-host, installed-runtime, and publication proof
  distinct.
- Preserve atomic old-or-new publication, pinned readers, crash recovery,
  content-verified freshness, fail-closed evidence, observational status/doctor,
  accelerated embedding without CPU fallback, proof-owned cleanup, and serial
  Cargo execution.
- Do not touch product version or release surfaces.
- The resolver and plugin manager must retain the production ban on spawning
  Git for repository metadata.

## Mutation contract

- Pin the accepted directory with handle-relative/no-follow operations and
  refuse terminal symlinks, reparse points, non-regular targets, and hard-linked
  aliases. A path replacement between inspection and open must not escape the
  accepted root.
- Preflight the complete hook trio before changing any file. One incompatible
  target means zero changes to all three.
[...]

PR #1736, opened 2026-08-02
[v0.17 CI-LANES] Run merged proof suites in per-PR CI lanes

TheGreenCedar:
Closes #1735.

## What

Adds one blocking step to `retrieval-engine-smoke.yml`'s universal `linux-contracts` job that runs the proof suites merged by EV-FIX (#1718), READY-A (#1727), HOOK (#1672), and W1.SEC (#1643) on every PR — 642 tests that previously executed only in the dispatch-gated workspace suite:

- `codestory-runtime --lib`: `agent::packet_evidence::` (9), `agent::packet_sufficiency::` (103), `agent::packet_batch::` (12), `tests::search_scoring_tests::` (41), `services::` (41)
- `codestory-cli --lib` (289, includes the stdio transport/catalog unit suites)
- `codestory-workspace` (147, includes the confined reader and repository-hooks boundary suites)

## Durability

The lane is pinned as a third `workflow_policy.proof_floor` entry (`merged_suite_lanes`) so it cannot be silently removed, reordered after artifact seeding, made conditional, or drift its command list:

- `release-claims.json` declares the lane; `scripts/codestory-release-claims.mjs` validates its exact shape (graph digest recomputed; all three fixtures updated).
- `.github/scripts/check-workflow-policy.mjs` enforces the step against the declared commands; five new mutation tests in `check-workflow-policy.test.mjs` prove removal/drift/conditionalization/nonblocking/reorder each fail policy.
- `docs/contributors/testing-matrix.md` documents the lane.

## Proof

- `node .github/scripts/check-workflow-policy.mjs` passes.
- `node --test .github/scripts/check-workflow-policy.test.mjs`: 1341 pass / 0 fail.
- `node --test scripts/tests/codestory-release-claims.test.mjs`: 40 pass / 0 fail.
- All seven lane commands verified non-empty (`--list`) and green locally at this head.

The seed-command topology digest is untouched, so cache keys are unchanged by construction.

### DIFF
--- a/.github/scripts/check-workflow-policy.mjs
+++ b/.github/scripts/check-workflow-policy.mjs
@@ -1387,6 +1387,36 @@ export function proofFloorPolicyViolations(
     `${architecture.workflow} architecture contract must complete before draft artifacts are seeded and saved`,
   );
 
+  const mergedLanes = object(policy.merged_suite_lanes);
+  add(
+    violations,
+    mergedLanes.workflow === retrievalFile
+      && mergedLanes.job === architecture.job,
+    `merged suite lanes must remain owned by the ${retrievalFile} universal job`,
+  );
+  const mergedLanesJob = object(at(
+    workflows.get(String(mergedLanes.workflow)),
+    "jobs",
+    mergedLanes.job,
+  ));
+  const mergedLanesStep = namedStep(mergedLanesJob, String(mergedLanes.step));
+  add(
+    violations,
+    sameStrings(
+      nonCommentLines(mergedLanesStep?.run),
+      list(mergedLanes.commands).map(String),
+    )
+      && mergedLanesStep?.if === undefined
+      && mergedLanesStep?.["continue-on-error"] === undefined,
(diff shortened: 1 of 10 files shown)


TheGreenCedar:
The lane immediately caught a latent platform-dependent test on its first CI run: `repository_identity::tests::bounded_logical_identity_fails_closed_when_a_missing_root_appears` asserted that a missing-then-appearing root changes the workspace id across observations. That only held on macOS, where tempdirs canonicalize through the `/var -> /private/var` symlink; on Linux `/tmp` is already canonical, both observations hash the same bytes, and the assertion fails. The suite had never run on Linux before this lane.

Fixed by dropping the two cross-observation id inequalities with an explanatory comment; the fail-closed contract — the per-observation-unique `Indeterminate` repository instance, `Absent` on the stable read, instance inequality, and convergence — is unchanged and still asserted. Round-1 failures (generalization-lint boundary in the claims loader, stale `candidate_sha256` fixture pin) were fixed in the previous commit.

PR #1810, opened 2026-08-03
[v0.17 S3-RUST] Extract Rust language rules behind the indexer registry

TheGreenCedar:
Closes #1684.

One rollback unit in the S3 series, following the pattern S3-KOTLIN (#1676) established.

## Proof

**Equality**: the two projection snapshots (fidelity lab, tictactoe) are byte-identical before and after the move. Goldens were captured from the pre-move tree, the move applied, then re-rendered and diffed — the committed goldens are literally the pre-move bytes.

**Mechanical**: every deleted line from `lib.rs` (and `language_configs.rs`) was matched against the moved bodies after undoing the declared renames — zero unmatched lines. Function accounting is stated in the commit body.

**Falsification**: registry fields were reverted and the snapshots observed failing, with the exact text recorded. Kotlin established why this matters — flipping `promotes_type_member_functions_to_methods` or changing `qualified_name_delimiter` fails these snapshots while `fidelity_regression` stays green, and for the delimiter no pre-existing test catches it at all.

No crate-root helper gained `pub(crate)` (private items are already visible to descendant modules), and the deliberate CLI-vs-route-scanner comment-roster disagreement was left alone.

### DIFF
--- a/crates/codestory-cli/src/output.rs
+++ b/crates/codestory-cli/src/output.rs
@@ -3941,8 +3941,8 @@ fn ansi_highlight_line(language: &str, line: &str) -> String {
     let comment_marker = codestory_contracts::language_support::line_comment_for_language(language)
         .or(match language {
             "bash" | "python" | "ruby" | "toml" | "yaml" => Some("#"),
-            "rust" | "typescript" | "tsx" | "javascript" | "jsx" | "go" | "java" | "csharp"
-            | "cpp" | "dart" | "php" | "swift" => Some("//"),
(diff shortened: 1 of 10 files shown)


TheGreenCedar:
Superseded by #1818, which lands all fourteen remaining S3 packages together.

They could not land separately: each was written against a `lib.rs` where the other thirteen had not moved, so every merge invalidated the next. Three collisions only exist once they are combined — `go`'s new `member_edge_specs` field, `typescript` repointing JavaScript call sites that `javascript` had already moved out, and five residual `match` seams that empty out once every language is registry-backed.

This branch's commit is preserved verbatim in #1818, so a single language can still be reverted on its own.

Issue #1686, opened 2026-08-02
[v0.17 S3-RUBY] Extract Ruby language rules behind the indexer registry


TheGreenCedar:
Package **S3-RUBY** in the v0.17.0 consolidated remediation program.

## Contract

Move only the Ruby extraction rules into indexer/src/languages/<lang>.rs, register template/style extensions through the language registry, and collapse Ruby-relevant name-roster maps onto registry iteration. Do not mix another language into this rollback unit.

## Dependencies

The preceding S3 language package (Go). Kotlin remains the first S3 migration; the declared sequence is Kotlin -> Java -> C++ -> C -> JavaScript -> TypeScript -> TSX -> Python -> Rust -> Go -> Ruby -> PHP -> C# -> Swift -> Dart -> Bash.

## Program boundaries

- The implementation PR uses a `codex/*` branch, targets `dev/codestory-next`, and carries this issue as `Closes #N`.
- Keep source, package, protected-host, installed-runtime, and publication proof distinct.
- Preserve atomic old-or-new publication, pinned readers, crash recovery, content-verified freshness, fail-closed evidence, observational status/doctor, accelerated embedding without CPU fallback, proof-owned cleanup, and serial Cargo execution.
- Do not touch product version surfaces. The release lane owns the single synchronized version change after implementation is integrated.
- Run focused proof while iterating and the owning path gate on the accepted candidate; broad source proof remains once on the frozen release candidate.

## Done when

Ruby fidelity_regression and tictactoe_language_coverage outputs are equal before and after the move, the architecture contract preserves ownership, and no behavior or packet_search_eval baseline changes.

Refs #1243
Refs #1179

Issue #1801, opened 2026-08-03
[v0.17 IDX-SPLIT] Split the indexer god file before it crosses its own oversized-source cap


TheGreenCedar:
## Context

`crates/codestory-indexer/src/lib.rs` is **986,987 bytes against the 1,000,000-byte oversized-source cap the indexer itself enforces** — 13 KB of headroom on a 28,438-line file.

`crates/codestory-indexer/tests/integration.rs::test_indexing_real_repo_rust_files_does_not_drop_projection_on_graph_error` copies this crate's own Rust sources into a temp repo and asserts they index with zero errors. The moment `lib.rs` crosses the cap that test fails with `Skipped oversized source file ...: N bytes exceeds 1000000 byte cap` — naming a temp path rather than the real file, which makes it slow to diagnose.

This already happened on `codex/1773-src-c2` (#1773) and was worked around by extracting tests into `crates/codestory-indexer/src/projection_fence_tests.rs`. With ~13 KB left, almost any further addition reproduces it.

## Seams that already exist in the file

- The callable / file-structural projection machinery: `build_callable_projection_states`, `structural_projection_fence`, `CallableProjectionExtents`, `classify_projection_update`, `ProjectionUpdateMode`.
- The per-language collectors (shrinking fast as S3 lands, but the residual dispatch remains).
- The trailing `#[cfg(test)] mod tests` block — **6,985 lines, 245,803 bytes, 25% of the file**. Largest single seam by a wide margin.

## Sequencing

The S3 language fleet (#1677–#1691) is removing an extrapolated **~7,860 lines (~269 KB)** from this file; eight extractions already measured at −4,192 lines. That takes `lib.rs` to roughly 718 KB on its own. Splitting while fifteen branches have the file open would turn one clean refactor into fifteen modify/delete conflicts, so the split lands **immediately after the fleet merges**, not before.

The pressure is relieved by the fleet; the structural problem is not. A 28k-line file behind a cap its own crate enforces stays fragile at 718 KB.

## Contract

- `lib.rs` splits along the seams above with every item's visibility unchanged, so no public surface moves.
- A guard fails with an actionable message *before* the cap is reached, naming the real file, rather than surfacing as an oversized-source skip on a temp path.

## Done when

- `cargo test -p codestory-indexer` (integration suite included), `cargo fmt --check` and `cargo clippy --all-targets` are green, and no crate source sits within 10% of the cap.

Refs #1243. Refs #1179.

PR #1818, opened 2026-08-03
S3: land the remaining fourteen language extraction packages

TheGreenCedar:
Lands the fourteen remaining S3 language packages together, one commit each so
any single language can still be reverted on its own, plus two integration
commits for the work no individual branch could contain.

Closes #1678, closes #1679, closes #1680, closes #1681, closes #1682,
closes #1683, closes #1684, closes #1685, closes #1686, closes #1687,
closes #1688, closes #1689, closes #1690, closes #1691.

## Why they had to land together

Each package was written against a `lib.rs` where the other thirteen had not
moved. Three kinds of collision only exist once they are combined:

- **A schema change.** `go` added `member_edge_specs` to `LanguageExtraction`.
  `ruby` and `csharp` needed real collectors, not `None` — the dispatcher
  consults the registry *before* its residual `match`, so their rows make the
  old arms unreachable and `None` would have dropped their MEMBER edges with
  nothing failing. `php` had written the collector but never wired the field.
- **A move that outran its caller.** `typescript` repointed two JavaScript call
  sites in `lib.rs`; `javascript` had already lifted those functions into its
  own module. The repointing belongs at the new home.
- **Emptied residuals.** With all fifteen languages registry-backed, five
  "registry first, residual match second" seams have no reachable arms left.

## Test claims restated

The per-branch tests asserted which languages had migrated *so far* — `tsx`
asserted `typescript` had no registry row; the comment-marker tests
accumulated contradictory expectations for `typescript`, `tsx` and `dart`.
Restating them against the finished state surfaced a real defect: `tsx` does
have a comment row, so the CLI must read its marker from the registry like
every other migrated language. Only `jsx` still answers from the local roster.

## A pre-existing failure fixed on the way

`moved_language_rules_keep_the_tictactoe_projection_byte_identical` was
already red on `dev`. `af20ff96` changed the file-structural identity format
while the S3 packages recorded their goldens against the old one — parallel
branches, each green alone. Only that one field moves in the goldens; every
node, edge and occurrence line is unchanged, and so is `body_hash`, which is
what distinguishes a format change from a projection regression.

## Verification

`cargo test --workspace`, `cargo clippy --workspace --all-targets`,
`cargo fmt --check` and `scripts/lint-retrieval-generalization.mjs` all clean.
The byte-identical projection suite passes for all sixteen registry rows,
which is the equality proof this refactor rests on.

`lib.rs`: 994,162 → ~770,000 bytes, which also unblocks #1801.


TheGreenCedar:
**Correction: this merged as a squash, so the per-language revert granularity is _not_ in `dev`'s history.**

The PR body and the closing comments on #1804–#1817 all said a single language could still be reverted on its own. That was the point of keeping fourteen separate commits, and squashing them defeats it. My mistake, and worth stating plainly rather than leaving the earlier claim standing.

What is actually true now:

- `dev/codestory-next` has one commit, `37747643`, containing all fourteen languages plus the two integration commits.
- The branch `codex/s3-remaining-14` is **not deleted** and still carries all sixteen commits (`e964b60f` cpp … `a96dca49` bash, then `7ad4e6a7` and `ce030978`). The per-language diffs are intact and reachable.

So backing out one language is still possible, just not a one-liner: take that language's commit from the branch and revert its diff against `dev`, e.g.

```bash
git revert --no-commit -m 1 $(git rev-parse origin/codex/s3-remaining-14~N)
```

or more practically `git show <lang-commit> | git apply -R`, then re-run the byte-identical projection suite, which is what would catch a partial back-out.

Please keep `codex/s3-remaining-14` around until v0.17.0 ships; deleting it is what would make this irreversible.

Issue #1845, opened 2026-08-08
[v0.17 READY-Bb] Rescope the freshness memo to the ReadyLease and enforce one fingerprint pass


TheGreenCedar:
## Context

READY-B slice after merged PR #1779 (child #1778), which declared this remainder. At head `383ed738`, `crates/codestory-workspace/src/source_freshness.rs` still documents an operation-scoped memo, while the readiness authority is the `ReadyLease` (`crates/codestory-runtime/src/services.rs:313`). Consequence, recorded at `crates/codestory-contracts/src/api/dto.rs:2655-2660`: a warm packet was measured performing SIX readiness fingerprint passes instead of one, with an explicit "unproven until enforced" note.

This slice is W3.3 plus the headline pass-count enforcement. W3.4 (publication-keyed canonical tables) is already live in `retrieval_primary.rs` — do not redo it. W3.6 (status/doctor lease evidence) is deliberately deferred until the S2 architecture-contract churn settles.

## Scope

- Rescope the source-freshness memo from operation lifetime to ReadyLease lifetime without weakening the same-mtime drift guard or the torn-read guard.
- Reduce warm-packet readiness fingerprint passes to one per lease, and enforce it: a test that counts passes on a warm packet and fails on more than one.
- Update the dto.rs documentation note to state the enforced truth.

## Out of scope

- W3.6 status/doctor changes; anything in `crates/codestory-cli`.
- `crates/codestory-runtime/src/index_freshness.rs` cap values and observer escalation (EV-7b #1651 owns them; it lands after this slice on the shared services.rs seam).
- stdio transport or status wire goldens.

## Done when

- Oracle-first: pre-change fixtures pin freshness verdict behavior for same-mtime drift, torn reads, and cross-operation reuse before the rescope; the moved memo reproduces them exactly.
- A warm packet performs exactly one readiness fingerprint pass per lease, enforced by test.
- Focused workspace + runtime suites, all-target clippy with warnings denied, and formatting pass.
- Independent fresh-context adversarial review accepts the exact head.

Refs #1700
Refs #1179

TheGreenCedar:
READY-Bb is merged. PR #1852 landed as merge commit `7b89bb70890b050829ef7c612c41bb25102ae925` (parents `3a91be0f`, `c302d26a`). Done-when evidence:

[...]

Issue #1700, opened 2026-08-02
[v0.17 READY-B] Make readiness verdicts lease-scoped and observational


TheGreenCedar:
Source package **READY-B** in the v0.17.0 consolidated remediation program.

## Contract

Deliver W3.3, W3.4, and W3.6: compute strict readiness once per `ReadyLease`; memoize operation-scoped freshness without weakening same-mtime drift or torn-read guards; cache the canonical-table map by publication; and make status/doctor consume existing lease evidence without scanning, observing, migrating, or repairing.

## Parent proof tracker

#1396 remains the protected installed-latency and live-readiness proof tracker. This child owns the remaining source implementation and its deterministic source gates; its PR closes only this child.

## Dependencies

READY-A / #1599 and the source-identity order declared by #1179.

## Program boundaries

- The implementation PR uses a `codex/*` branch, targets `dev/codestory-next`, and carries `Closes #N` for this child only.
- Status and doctor remain observational. A source cache or responsive process is not installed-runtime proof.
- Preserve content-verified freshness and fail closed on identity or receipt drift.
- Do not touch product version surfaces.

## Done when

A warm packet performs exactly one readiness fingerprint pass; same-mtime mutation remains detectable; canonical-table work is publication-keyed; and hostile tests prove status/doctor cannot create observers, scan, migrate, or repair. #1396 stays open for the protected installed measurements.

Refs #1179

TheGreenCedar:
READY-B is complete; closure evidence across its three slices: W3.3 + the enforced one-fingerprint-pass headline landed in PR #1852 (lease-scoped freshness memo, drift/torn-read guards adversarially proven live, the six-pass warm packet reduced to an enforced 1); W3.4 (publication-keyed canonical tables) was verified already live at audit time; W3.6 landed in PR #1864 (status/doctor consume lease evidence additively with hostile observational proof). The workspace/lib.rs chain ordering the epic mandated was honored, and proof tracker #1396 retains the installed-latency measurement against the frozen candidate. Follow-ups recorded: #1857 (memo guard tests + a None-after-Some ordering scenario for the hostile matrix), #1853 (flaky macOS incremental-publication fixture).

PR #1852, opened 2026-08-08
[v0.17 READY-Bb] Scope the freshness memo to the ready lease, one fingerprint pass

TheGreenCedar:
## Context

READY-Bb, the W3.3 remainder of #1700 declared by PR #1779. The source-freshness memo was operation-scoped while the readiness authority is the `ReadyLease`; consequence, recorded in `dto.rs`, was a warm packet paying SIX whole-repository readiness fingerprint passes with an explicit "unproven until enforced" note.

Closes #1845
Refs #1700
Refs #1179

## What changed

- `source_freshness.rs`: the verdict memo becomes an `Arc`-shared `SourceFreshnessMemo` a runtime ReadyLease can own (`enter_with_memo`); plain `enter()` keeps the exact prior per-operation semantics (fresh private memo, inert at depth 0). `reverify_from_content` still drops every verdict, so drift checks answer from disk. New lease-scoped opaque-value surface (`with_lease_memoized_value`, `invalidate_lease_memoized_values`) with invalidation wired to freshness refusals.
- `services.rs`: the activation service arms the lease-owned memo per target; freshness refusals invalidate lease-scoped observations at both the admission check and the post-build reverified check.
- `index.rs` (retrieval): `compute_sidecar_input_fingerprint` reuses a lease-memoized fingerprint keyed on the full identity tuple; the computing path is unchanged and retrieval learns nothing about leases — it consumes the opaque-value surface workspace exports.
- `retrieval_primary.rs`: active retrieval pins bypass the five redundant readiness gates (minimal, local diff; `CachedCanonicalSymbolNames` untouched).
- `dto.rs`: the pass-count note now states the enforced truth.

## Verification

Passed on `3046aba9`:

- One-pass enforcement: first warm packet = 1 fingerprint pass, same lease = 0, replacement lease = 1; a temporary mutation disabling the precomputed handoff failed the test with `left: 2, right: 1` and was reverted.
- Oracles: same-mtime/same-length drift detection, deterministic torn-read refusal, same-lease reuse / new-lease recomputation — all pinned and green.
- Host suites: `codestory-workspace` green, `codestory-retrieval --lib` 292 passed, `codestory-runtime --lib` 1088 passed with exactly one failure — `incremental_publication_ignores_changed_files_without_graph_collectors`, which fails identically at the base `383ed738` on macOS (bisect-verified with the full diff stashed); it is pre-existing and environment-local, tracked separately.
- Three-crate clippy `-D warnings`, `cargo fmt --check`, `git diff --check` all pass.

Independent fresh-context adversarial review on the exact head is running; draft until it accepts and CI is green.

### DIFF
--- a/crates/codestory-contracts/src/api/dto.rs
+++ b/crates/codestory-contracts/src/api/dto.rs
@@ -2650,13 +2650,12 @@ pub struct PacketClaimProfileTelemetryDto {
 pub struct SourceFreshnessTelemetryDto {
     /// Stored files whose content was read and hashed for a freshness verdict.
     pub content_hash_reads: u32,
-    /// Freshness verdicts served from the operation-scoped verdict memo.
+    /// Freshness verdicts served from the ready-lease-scoped verdict memo.
     pub verdict_reuses: u32,
     /// Strict-readiness fingerprint passes. Each reads the repository's
     /// lexical source live off disk and streams both projection tables. This
-    /// is a measurement, not a bound: a warm packet over an unchanged
-    /// repository was measured at six passes, so treat any single-pass
-    /// expectation as unproven until it is enforced.
+    /// is a measurement, not a bound: the first warm packet on one ready lease
+    /// performs exactly one pass, and later operations reuse that fingerprint.
     pub readiness_fingerprint_passes: u32,
 }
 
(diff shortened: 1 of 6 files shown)


TheGreenCedar:
Head moved to c302d26a: the only delta from the adversarially ACCEPTED 3046aba9 is removal of the retrieval_primary.rs pin short-circuit — the exact hunk the verifier's M4 mutation removed and re-tested (full runtime suite + the ==1 enforcement test green without it, and it was the slice's one observable-verdict change). Supervisor re-ran the one-pass test, runtime clippy -D warnings, and fmt on c302d26a. The slice is now strictly behavior-preserving; acceptance carries via the verifier's own M4 execution plus these re-runs. Merging on green CI.

Issue #1855, opened 2026-08-08
[v0.17 REL-MANb] Executable catalog rollback, recovery semantics, and candidate-pinned catalog proof


TheGreenCedar:
## Context

REL-MAN slice after merged PR #1796, which delivered the lane split, release manifest, launcher consumption, and W8.6a discharge. The audited remainder of #1671 at head `3a91be0f`: automatic prior-SHA catalog restore on live-catalog smoke failure, the recorded "release incomplete until catalog recovery succeeds" state, the `workflow_policy.catalog_delivery` semantics that #1671 explicitly says must change before automatic rollback/recovery can be claimed (`recovery.automatic_restore` is still `false`), and the W8.4 candidate-pinned pre-publication marketplace catalog proof. Ed25519 arming stays a recorded v0.17 non-claim per the epic ("digest-only containment is stated until signatures arm" is already satisfied in `docs/architecture/host-integration.md`).

## Scope

- Automatic prior-SHA catalog restore: when post-publish live-catalog smoke fails against a just-pushed catalog, the recovery path restores the recorded previous catalog SHA rather than leaving a broken live catalog, reusing the marketplace-sync credential path; restore is idempotent and its outcome is a first-class recorded state.
- Extend the catalog-delivery identity registry coherently: any new state lands in `.github/scripts/marketplace-delivery-identity.mjs`, the `marketplace_installation.py` predicate, and `release-claims.json` together, per the standing three-name rule.
- Record the honest ledger state: a release whose catalog delivery failed and was restored (or deferred) is complete as a release but carries "catalog sync deferred/restored" — the release itself must still never fail on catalog delivery. `release_gate` semantics change only as far as #1671's contract authorizes: rollback/recovery becomes claimable because it is executable, not because catalog delivery starts gating the release.
- W8.4: a candidate-pinned pre-publication marketplace catalog proof — prove the install path against the candidate-pinned fixture before publication, without requiring the live catalog to match an unreleased commit.
- Update `check-workflow-policy.mjs`/tests, claims fixtures, and the runbook/docs wherever they describe the recovery story.

## Out of scope

- Ed25519 machinery (recorded non-claim), native-lane frozen-head choreography wiring (REL-CI owns it), product version surfaces, any Rust.

## Done when

[...]

Issue #1671, opened 2026-08-02
[v0.17 REL-MAN] Bind native release manifests and recover catalog delivery


TheGreenCedar:
Package **REL-MAN** in the v0.17.0 consolidated remediation program.

## Contract

Deliver W8.4, W8.6b, and W8.11 as the publication unit: candidate-pinned marketplace proof and previous catalog SHA, catalog recovery semantics in machine policy, a release-generated manifest carrying target/archive length/SHA-256 and release identity, launcher consumption before extraction, and optional Ed25519 arming with rotation/revocation and no unsigned fallback once armed.

Discharge W8.6a's intentionally provisional source-pin assertion by lane:

- plugin-source provision proof continues to require the pinned archive digest already carried by the source pin;
- native frozen-head/pre-publish provision proof consumes the staged release manifest that owns the native archive digest, or defers that digest assertion to the post-release manifest proof;
- the native lane must never run the current unconditional source-pin archive assertion, because a lawful native pin has no archive digest.

## Dependencies

REL-CI and W8-A. Key ceremony is not a v0.17 blocker; current release_gate:false semantics must change before automatic rollback/recovery can be claimed.

## Program boundaries

- The implementation PR uses a `codex/*` branch, targets `dev/codestory-next`, and carries this issue as `Closes #N`.
- Keep source, package, protected-host, installed-runtime, and publication proof distinct.
- Preserve atomic old-or-new publication, pinned readers, crash recovery, content-verified freshness, fail-closed evidence, observational status/doctor, accelerated embedding without CPU fallback, proof-owned cleanup, and serial Cargo execution.
- Do not touch product version surfaces. The release lane owns the single synchronized version change after implementation is integrated.
- Run focused proof while iterating and the owning path gate on the accepted candidate; broad source proof remains once on the frozen release candidate.

## Done when

The frozen source contains no circular native digest; release-built bytes are manifest-bound; plugin and native provision proof each assert the digest source their lane actually owns; the frozen native head cannot fail solely because its lawful source pin lacks archive digests; digest-only containment is stated until signatures arm; catalog published/deferred/recovered identities remain distinguishable and rollback behavior is executable before docs claim it.

Refs #1233
Refs #1179

TheGreenCedar:
REL-MAN is complete; closing against the Done-when, clause by clause:

[...]

Issue #1233, opened 2026-07-17
[v0.17 proof] Close every declared release claim on one exact evidence ledger


TheGreenCedar:
Parent: #1189

## Outcome

Require one exact-head ledger chain. The accepted pre-publish ledger closes
every graph-required pre-publish cell before publication; the cumulative
post-publish ledger binds the shipped assets and catalog-installed runtime and
closes every remaining graph-required cell afterward.

## Context

The claim graph declares release claims, but the live path does not yet prove
that one exact source/tree and the complete package, installed-runtime,
protected-hardware, and post-publish identity matrix were evaluated together.
Workflow policy can prove structure; it cannot substitute for retained runtime
evidence.

The per-user embedding server adds distinct source and installed-runtime cells.
The ledger must retain both rather than accepting source-only owner tests.

## Current state

#1305 is closed and integrated through PR #1307 at merge commit `4cf779f1399d6f5f616b16a85358f49975769a5c`. The production workflow now contains the authenticated proof-cell wiring; source integration is not final release evidence.

Execute the ledger exactly once after F is frozen. The remaining acceptance is the graph-derived 10-cell pre-publish ledger and cumulative 22-cell post-publish ledger for one repository/commit/tree/version identity: 1 exact-source + 3 package + 3 accelerator-execution + 3 candidate-installed cells before publication, then 3 retrieval-readiness + 3 platform-support + 3 catalog-resolved installed-runtime + 3 downloaded-byte-equality cells after publication. Every cell retains the declared artifact/run/attempt/job/container and applicable host/runtime/native identity; no required cell is waived.

## Acceptance criteria

- [ ] Add exact-source proof to the release chain and bind every proof manifest
      to repository, commit, source tree, version, artifact digest, target,
      host, and applicable runtime/native identities.
- [ ] Emit one machine-readable manifest for every required graph-owned
      package, platform, installed-runtime, protected-accelerator,
      retrieval-readiness, and shared exact-source cell. Retain optional
      performance and answer-quality evaluation records separately under their
      graph-declared nonblocking/unclaimed contracts; they are not closeout
      cells and cannot replace one.
- [ ] Require exact #1220 source identity plus #1221 installed two-session,
      one-server/one-model behavior and the explicit frozen-server non-claim.
- [ ] Require the accepted PR #1262 Windows package/staging identity and #1230 trusted
      workflow/run/cache/reviewed-SHA identity.
[...]

PR #1860, opened 2026-08-08
[v0.17 REL-MANb] Executable catalog rollback with a recorded restored state

TheGreenCedar:
## Context

REL-MANb, the audited remainder of #1671 after PR #1796. Catalog rollback existed as prose and a manual sync path; the machine policy honestly recorded `automatic_restore: false`. #1671's contract requires rollback to become executable and recorded before any doc claims it, with the published/deferred/recovered identities distinguishable end to end.

Closes #1855
Refs #1671
Refs #1233
Refs #1179

## What changed

- **Automatic prior-pin restore**: when the post-publish live-catalog smoke fails against the catalog a release just pushed, the recovery path restores the recorded previous catalog pin, fenced by the pushed catalog revision, SHA, and version so it can never clobber an unrelated later push. Idempotent; non-rollbackable catalogs keep their existing refusal.
- **New `restored` closeout state**, coherent across all three registry surfaces per the standing rule: `marketplace-delivery-identity.mjs` (installer `codex_marketplace_restored_fixture`), `marketplace_installation.py` (repository `local:candidate-pinned-marketplace-restored-fixture` accepted shape), and `release-claims.json`. Restored and deferred releases resolve candidate-pinned fixture proof and can never claim `published`.
- **Semantics**: `workflow_policy.catalog_delivery.recovery.automatic_restore` is now `true` — together with the machinery that makes it true. `release_gate` remains `false`: catalog delivery still never fails a release; what changed is that recovery is executable and its outcome is a first-class recorded state.
- **W8.4**: candidate-pinned pre-publication marketplace catalog proof in the plugin lane; the native lane's existing proof posture is unchanged.
- Claims fixtures regenerated coherently including the attested `candidate_sha256`; runbook updated only where behavior is now test-proven.

## Verification

Passed on `2835c274` (implementer where the sandbox allowed; every suite re-run by supervisor with full deps): publish-catalog tests 3/0, policy checker, policy tests 1369/0, claims 43/0, closeout 39/0, evidence-gate 20/0, plugin-static 130/0, Python marketplace-delivery self-tests, actionlint + its tests, doc links, generalization lint, `git diff --check`.

Independent fresh-context adversarial review on the exact head is running; draft until it accepts and CI is green.

### DIFF
--- a/.github/scripts/check-workflow-policy.mjs
+++ b/.github/scripts/check-workflow-policy.mjs
@@ -2902,6 +2902,14 @@ function validatePluginAndDraftWorkflows(workflows, violations, graph) {
 // lets a run that never touched the catalog report that it did.
 function catalogDeliveryOutcomeViolations(file, job, delivery) {
   const violations = [];
+  const checkout = list(job.steps).find(
+    (step) => String(object(step).uses ?? "").startsWith("actions/checkout@"),
+  );
+  add(
+    violations,
+    object(object(checkout).with)["fetch-depth"] === 0,
+    `${file} marketplace publication must retain history needed to restore the recorded prior SHA`,
+  );
   const tokenStep = namedStep(job, "Mint a scoped marketplace token");
   add(
     violations,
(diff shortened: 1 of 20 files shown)


TheGreenCedar:
Head moved to c7cb73e25fd06598c0992e8e4ef40d8aa2a0a95e: the only delta from the adversarially ACCEPTED 2835c274 is one added test killing the single retained defect (M2) — the restore-fence refusal now has behavioral coverage; the fence-deletion mutant fails it with the clobbered-catalog evidence, exactly the verifier's race experiment codified. Publisher suite 4/0 (supervisor re-run). Merging on green CI.

Issue #1925, opened 2026-08-17
Make packet evidence selection obligation-aware


TheGreenCedar:
## Objective

Improve packet evidence quality for material flow questions by selecting verified source ranges and typed graph receipts against stable semantic proof obligations, without query-token heuristics, repository-specific steering, higher global caps, or a material latency/reliability regression.

## Current evidence

- Monolog retrieves the relevant logging code, but packet compilation can omit the handler iteration/dispatch source even when an exact typed CALL receipt exists.
- AutoMapper can retrieve `TypeMap.cs`, then lose it during bounded candidate resolution.
- Animate can retrieve its source entrypoint, variables, and animation files, then lose them before final packet selection; CSS import targets and class-to-keyframe linkage also need truthful typed receipts.
- Jekyll is the unaffected regression guard.
- Two selector-shaped revisions and one producer-plus-selector revision were rejected and reverted. The next design must move proof authority out of name-shaped carrier predicates.

## Required architecture

- Retrieval query IDs remain provenance only.
- Material obligations use typed proof atoms over exact verified source aspects and typed relation receipts.
- Candidate admission, proof verification, and final selection remain separate.
- The selector reuses existing retrieval, graph, source-read, caps, one-drill, and fail-closed machinery.
- No benchmark vocabulary, prompt-token proof heuristics, quoted-path reconstruction, global top-k changes, ranking-weight tuning, or embedding changes.

## Acceptance

- Direct exact-binary packets retain the required Monolog, AutoMapper, and Animate evidence; Jekyll does not regress.
- Two exact-binary Monolog agent repeats plus one AutoMapper, Animate, and Jekyll run are at least as accurate as the unchanged baseline while materially reducing agent work.
- No material packet-latency or reliability regression.
- Focused serial `--locked` checks pass.

Refs #1179

PR #1926, opened 2026-08-17
Improve packet evidence selection quality

TheGreenCedar:
## Status

Draft quality lane based on the proven 0.17.0 checkpoint now on `dev/codestory-next`. The branch currently contains four accepted bounded improvements; obligation-aware proof/selection work remains in research and is not yet implemented.

## Current source

- branch head: `ccebbe622decb84b20b5c04dd66a26ed4f9a150e`
- release checkpoint ancestor: `37c437607ec1ac77396770118a58c3235df3cc83`
- dev checkpoint merge: `8173044227c012b95a5d409037f14a5e27826ca8`

## Already on this branch

- preserve proven connected flow carriers
- bound concurrent retrieval fanout
- preserve adaptive owner identities
- tighten packet retrieval anchors

## Next architecture

Move semantic proof authority from name-shaped citation predicates to typed proof atoms over verified source aspects and typed graph receipts. Retrieval queries remain provenance only. Reuse existing retrieval, graph, source-read, cap, drill, and fail-closed machinery.

Do not add repository vocabulary, prompt-token proof heuristics, quoted-path reconstruction, global top-k/ranking changes, or embedding changes.

## Gate

Direct exact-binary packets must pass Monolog, AutoMapper, Animate, and the Jekyll guard before any exact-agent A/B. Reuse unchanged baseline rows. Commit only a candidate that improves the affected exact-agent shard without a material latency or reliability regression.

Closes #1925
Refs #1179


TheGreenCedar:
**Checkpoint: stage-0 proof-formula contract accepted** (head unchanged at `ccebbe62` — design phase, no code yet).

The typed-proof implementation now has an adversarially reviewed design contract: `autoresearch.research/obligation-evidence-selector/2026-08-17-stage0-proof-formulas.md` (revision 5, untracked research artifact). Five review rounds; the final pass confirmed every atom's discharge inputs trace to receipt fields that exist or to named deltas at verified seams. Key properties: all cross-receipt joins are node-identity joins (segment values demoted to single-receipt classification), containment never reads prompt-chosen window extents, certainty is attributed per edge kind to the component that actually sets it, and the vocabulary-gate demotion (R1) is specified at all five function sites where it is load-bearing plus one stated exemption.

Notable contract decisions vs. the original handoff sketch:
- PHP construction receipts (`syntax:php-new` constructor CALLs) are **required**, not optional — `logger_event` is not provable by source reread alone.
- The candidate-resolution boundary gets an identity-only re-prioritizable admission queue (R6) — resolution budget preserved exactly, base order demoted to default; this is where the passive ledger measured the CSS evidence dying (ranks 71/338 vs a ≤50 window).
- The CSS base stylesheet is characterized structurally (selector uses the vars-file's custom property AND has no keyframe-binding USAGE, truncation-guarded) — verified against the real animate tree that this uniquely selects `_base.css` against all six var-using sibling files.

Implementation started: stage 1 (pure proof-atom module in codestory-agent) and stage 3a (PHP producer: scope primitive, PHPDoc reader, foreach dispatch, construction) are in flight as isolated lanes. Next proof target: stage-1 module green + F1 adversarial review, then receipt plumbing (stage 2) with the Jekyll-only direct-packet gate.

TheGreenCedar:
**Checkpoint: stages 1–2 landed** — head `e4e9ea06`.

- `00e12870` adds the pure proof-atom model and matcher (leaf module in codestory-agent, 43 tests, leaf-ness asserted by the architecture contracts). CI fully green including both full-agent-suite lanes.
- `e4e9ea06` threads verified receipts through both obligation-finalization sites with one shared strict-subset filter, routes the six shard requirements to the matcher (everything else explicitly Legacy, bit-identical), adds the post-compile fail-closed reconciliation (single recompile, sufficiency proven by idempotence + plan-independence tests), and keeps proof verdicts in the env-gated step trace only.

Both stages passed adversarial review (F1: two rounds — an unsound absence-scan arm and a per-requirement API gap fixed; F2: one round — a name-token predicate could still select the public status level on the non-Proved path, now always Unsupported). **Jekyll guard verified on the exact benchmark prompt: disposition `supported`, both site obligations `proven` via Legacy.** Expected transitional state: the three affected-shard packets fail closed under honest proof authority until the producers (stage 3, in review: PHP accepted with the Monolog oracle exact; C# in round 2 after the AutoMapper oracle exposed primary-constructor/chained-call/csproj-using gaps) and the admission/protection deltas (stage 4) land.

Process note for future gates: the harness exits 0 even when cache prep fails — verdicts are read from rows and direct packet inspection, never exit codes; and the embedded-model build uses `CODESTORY_EMBED_MODEL_SOURCE` via `scripts/prepare-embedded-model.mjs` (the handoff's `--features embedded-model` flag does not exist).

TheGreenCedar:
**Checkpoint: stage 3 producer wave landed** — head `cae0114d` (30 files, +6658/−991).

Three language producers, each accepted only after its real-target index oracle matched exactly (fixture suites alone were provably insufficient — every wave failed its first oracle on a shape the fixtures missed, and the misses were all general-language semantics, not benchmark tuning):

- **PHP** (Monolog oracle): the dispatch edge at Logger.php:391 reproduces the run-6 reference (certain, receiver-owner + `receiver-binding:loop-element@369-399` with the callsite inside the range); constructions resolve to `__construct` including **multi-line named-argument, nested, and same-namespace-unqualified** shapes (first-oracle miss). Unrelated-edge counts invariant vs an unpatched baseline; certain construction receipts 79→479 after the same-namespace arm.
- **C#** (AutoMapper oracle): `TypeMapPlanBuilder→TypeMap` TYPE_USAGE certain from a **C#12 primary-constructor parameter on a ref struct**, and the execution-plan handoff CALL at TypeMap.cs:282 (**chained call on an object creation** in an expression-bodied method) resolves certain — both first-oracle misses, plus **csproj-level `<Using>` directives** handled via an engine-side same-root-namespace unique-declaration arm (ambiguity fails closed). TYPE_USAGE certainty is established producer-side (emit or pre-publication finalize; pending/uncertain edges never discharge — contract adjudication recorded).
- **CSS** (animate oracle): import statements are MODULE nodes with file-to-file IMPORT edges resolved to canonical file identities (99/99 on the entrypoint); keyframes are FUNCTION nodes with certain selector linkage; custom properties mint only at declarations (repo-wide VARIABLE count 15, down from noise); and the C3 discriminator is exact on the real tree (base selectors: zero keyframe usages; all six impostor files: nonzero). First-oracle miss: var() resolution needed the **bidirectional import-graph component** (sibling sheets under a shared importer), not the downstream closure.

Review trail in `.agent/review-003/004/006.md`. Next: stage 4 (admission queue, hydration/protection/anchored-verification deltas, bounded selection) is implementing; the full four-packet direct gate and F3 adversarial review follow before its commit.

TheGreenCedar:
## Decision point after gate round 6 — full picture and a recommendation

**Committed and green:** `00e12870` (proof-atom model + matcher), `e4e9ea06` (receipt plumbing through both finalize sites), `cae0114d` (three language producers). Each passed adversarial review and its own gate; all CI lanes green; `main` untouched.

**Uncommitted:** stage 4 (admission, retention, protection, anchored verification, selection) — ~2,000 production lines, nine files, independently verified at runtime 909 / agent 307 / contracts 44, F3-reviewed with its blocker fixed.

**Gate status (four-packet direct gate):**
- `php-monolog-record-flow` **PASS, stable** — the shard three prior candidates died on now proves end to end from real typed receipts.
- `ruby-jekyll-site-build` **PASS, stable** — the all-Legacy guard, provably untouched.
- `csharp-automapper-map-flow` **FAIL** (files 0.25, claims 1.0) — `mapper_config` proves for the first time ever, but on a generic binding; the true `TypeMapPlanBuilder`→`TypeMap` chain (edge verified present in the index) never gets admitted.
- `css-animate-base-and-keyframes` **FAIL** (files 0.6 — that leg passes; symbols 0.5 vs 0.55) — the C chain works and cites entrypoint + `_vars.css` + `_base.css`; it cites `fadeIn.css` where the manifest expects `bounce.css`/`flash.css`.

**Six rounds eliminated, in order:** trail fan-out burning the retrieval stage clock; per-query promotion state with no cross-query bootstrap; unfiltered promotion mass-displacing base evidence; need-set flooding from generic membership/usage patterns; **three environmental cache-provenance traps that made rounds 1–4's A/C numbers invalid measurements** (documented in `.agent/acceptance-protocol.md`); and a store-accessor zero-retention pathology. All real, all fixed or documented.

**What remains is formula discrimination, not a bug.** Admission priority by atom-role multiplicity is in flight and should close AutoMapper. For Animate, C4 treats every animation source as interchangeable while the manifest expects two specific ones — **admission cannot fix what the formula accepts**, and I deliberately declined to add an import-order preference to make the benchmark's two files win; that would be fitting the benchmark.

**Recommendation:** commit stage 4 to keep the reviewed work safe, keep this PR draft, and take the C-formula expressiveness question to a scoped contract review rather than a seventh admission round. Full analysis with options in `.agent/DECISION-POINT.md`.

**The exact-agent A/B has NOT been run and will not be until all four packets pass** — that gate is exactly why three prior candidates produced worthless numbers.

TheGreenCedar:
## Stage 4 landed as `003a110e` — with an honest gate statement, and iteration stopped at 8 rounds

**Committed** (draft PR only; `main` untouched, no release path): admission queue with atom-need-gated, cross-container-restricted, slot-bounded, multiplicity-prioritized promotion; off-stage-clock retention with narrowed coverage sets; atom-driven protection; anchored verification; selection. Plus a genuine matcher correctness fix: when a requirement group cannot prove as a whole, requirements are now attempted **most-constrained-first** rather than in declaration order, so a weakly constrained requirement can no longer capture a shared role and force its stronger sibling onto wrong bindings. Verified independently: agent `--lib` 312, runtime `--lib` 912, architecture contracts 44, fmt/diff clean.

**Gate status:** `php-monolog-record-flow` and `ruby-jekyll-site-build` pass, stably, across every round. `csharp-automapper-map-flow` (files 0.25) and `css-animate-base-and-keyframes` (files 0.6, symbols 0.5 vs 0.55) do not.

**Why I stopped at round 8:** gates 6, 7 and 8 produced *byte-identical* quality numbers under three genuinely different fixes. That invariance is itself the finding — the remaining loss is not in the seam I kept adjusting.

**The residual is now named precisely.** Trace on the final binary: `AutoMapper.TypeMap` and `AutoMapper.Execution.TypeMapPlanBuilder` are **both admitted and resolved** — six rounds of admission work do deliver the true chain — but neither is cited, and `citations` appears in the packet's `omitted_sections`: the citation cap drops them. Meanwhile `mapper_config` proves on a weaker binding and holds the protection that should have reserved the true chain, and `mapper_execution` cannot prove because its CALL+MEMBER receipts aren't retained for **late-admitted** candidates (MEMBER trails run post-pass — i.e. after the citation cap has already selected).

**Next seam, concretely:** (1) retain an atom's required typed edges for candidates admitted late in the pass, and (2) make partial-atom protection follow the *strongest available* binding rather than whichever requirement proved first.

**What I deliberately did not do:** add an import-order or file-position preference to make Animate's two expected files win. C4 accepts any imported keyframe-bearing source; the manifest expects two specific ones. That is a formula-expressiveness question for contract review — tuning admission around it would fit the benchmark and poison the acceptance signal.

**No agent A/B has been run** and none should be until all four packets pass. Full analysis: `.agent/DECISION-POINT.md`; round-by-round forensics: `.agent/review-008.md`; environmental trap checklist for acceptance: `.agent/acceptance-protocol.md`.

TheGreenCedar:
## `eb6a82de` — three of four bounded selections fixed, and one plain defect found along the way

I was wrong to stop at round 8. Pushing further found the actual mechanism, and it turned out to be four successive bounded selections where atom-needed evidence had to survive and only the first knew about it.

**The defect worth flagging on its own merits, unrelated to typed proof:** a deadline-cancelled sidecar *batch* query discarded every hit it had already resolved — 327 of them on AutoMapper — while the single-query path (`sidecar_primary_result_rejection_reason`) deliberately *serves* resolved hits from a cancelled query. That asymmetry costs real retrieval work on any deadline-pressured packet in the product. The batch path now matches its sibling: hits retained in resolution order, still retryable, with atom need only breaking ties among equal ranks — so packets carrying no formulas keep their previous hits exactly.

Also fixed: the citation carry and graph-edge cap now rank atom-required evidence ahead of unrelated evidence *within existing limits* (previously lexicographic edge ids decided which receipts survived); the byte trimmer shaves a non-atom edge when one exists but never treats atom edges as protected, so it can always still trim and `max_output_bytes` holds; post-pass hydration walks roots in need order and skips what it can't afford instead of stopping at the first unaffordable one.

**Gate 10, three repeats per packet** (the implementer caught that this shard is unstable run-to-run, so single-run comparisons were sampling noise):

| packet | files | symbols | verdict |
|---|---|---|---|
| `ruby-jekyll-site-build` | 1.000 ×3 | — | **pass ×3** |
| `php-monolog-record-flow` | 1.000 ×3 | 0.833 | **pass ×3** |
| `csharp-automapper-map-flow` | 0.500 ×3 | 0.667 ✓ | fails **only** file recall (0.60) |
| `css-animate-base-and-keyframes` | 0.600 ✓ ×3 | 0.500 | fails **only** symbol recall (0.55) |

The generalized retention also **stabilized** AutoMapper — it was swinging 3/4, 2/4, 2/4 on identical code and is now flat, with file recall doubled from gate 8.

**Baseline check:** a stage-3-only binary built in an isolated worktree also cancels 32/32 AutoMapper sidecar queries, so the cancellation storm is pre-existing product behaviour, not something this work introduced.

**Both remaining gaps are single-leg and both are formula discrimination, not selection:** 294 equally-admissible TYPE_USAGE identities compete for 16 file slots (atom-role multiplicity measurably does not separate them — both retention orders gave identical recall), and C4 accepts any imported keyframe-bearing source where the manifest expects two specific ones. I attempted neither A1 discrimination nor any file-position/import-order tiebreak; inventing one that happens to select the benchmark's expected files would be fitting, not fixing.

Record: `.agent/review-008.md` (round-by-round), `.agent/DECISION-POINT.md`, `.agent/acceptance-protocol.md`.

Issue #2148, opened 2026-09-09
Recognize published core generations in retrieval readiness


TheGreenCedar:
Fresh installed search rejects a valid generation-based retrieval publication as missing because readiness checks the absent legacy database file.

Reproduced on clean integrated candidate `975020e3d4ca35be0696ee895446bcc2e2426a0f` through the ordinary managed MCP launcher and native Metal execution. The core pointer, active generation and retrieval publication exist; the logical flat `codestory.db` does not. Earlier activation accidentally created that empty file and masked the error.

Use the store-owned core existence resolver in retrieval status. Preserve strict artifact, source, engine and manifest validation; absent or corrupt publications must fail closed without mutation. Add generation-only and legacy status coverage, then rerun the same fresh installed interaction.

Owner: root implementer and independent compatibility verifier. Branch `codex/0176-generation-readiness`, worktree `/Users/albert/.codex/worktrees/0176-generation-readiness/CodeStory`, base `975020e3d4ca35be0696ee895446bcc2e2426a0f`. No panel model sessions have run. Candidate qualification remains invalid until fixed.

Refs #2135.

PR #2149, opened 2026-09-09
recognize core generations in retrieval status

TheGreenCedar:
Fresh installed search can prepare retrieval from an immutable core generation without a leftover legacy database file. Previously readiness skipped a matching retrieval manifest when the logical `codestory.db` path was absent, even after successful native embedding. Strict readiness also reopened legacy source-policy storage through a reader that could create SQLite sidecars.

Existence now uses the store-owned publication resolver, and source-policy reads use its observational API. Existing source, artifact, manifest, policy and engine checks remain in place. Missing generations, corrupt pointers and pending promotion recovery fail closed. Review the two production changes in `sidecar.rs` and `lexical_index.rs`.

Both regressions failed before their respective fixes. The complete retrieval suite passed on final confirmation: 410 library and 6 integration tests, with 3 existing tests ignored. One earlier full run failed an unrelated SCIP cache-reuse assertion; its isolated diagnostic and the unchanged default-parallel suite passed. The original failure is retained; shared four-entry cache eviction is a source-based explanation, not independently instrumented proof. Independent production and installed acceptance passed on `7aa80fc6`: 16 hostile full/descriptor observations preserve cache paths and bytes; ordinary managed fresh, warm and restarted search reach full Metal retrieval without a legacy flat database. Damaged-pointer status rejects without cache mutation or native execution. The synthetic mismatched-target descriptor fixture lacks lexical artifacts, so it does not independently prove identity rejection with otherwise healthy artifacts. Linux CI also exposed a fixture that removed only the legacy file while labelling the still-valid generation as missing. The revised fixture checks generation-only readiness, then removes the actual publication pointer and proves unavailable status; all cache-preservation assertions remain unchanged. The complete runtime services suite passes all 61 tests. This final delta changes tests only. No source stabilization, package qualification or release acceptance is claimed.

Source `fce371208dd6bb772a5a0269c32468323abc6552`, tree `be34a04201d9b277ce5166fc6fac06a04af0cac9`, base `975020e3d4ca35be0696ee895446bcc2e2426a0f`. Worktree `/Users/albert/.codex/worktrees/0176-generation-readiness/CodeStory`. Root implements; the independent verifier owns the eight-state full/descriptor matrix and fresh installed search/restart. The final test-only delta was independently executed and accepted on `fce37120`; production and the shared observer assertions remain byte-identical to the accepted source. Required CI remains pending before merge; the final integration archive must repeat the same fresh installed search before the maintenance panel. All 36 maintenance model sessions remain unstarted.

Closes #2148. Refs #2135.

### DIFF
--- a/crates/codestory-retrieval/src/lexical_index.rs
+++ b/crates/codestory-retrieval/src/lexical_index.rs
@@ -3584,8 +3584,8 @@ fn lexical_source_policy(
             storage_path.display()
         );
     }
-    let storage =
-        Store::open_read_only(storage_path).context("open storage for lexical source policy")?;
+    let storage = Store::open_observational(storage_path)
+        .context("observe storage for lexical source policy")?;
     let publication = storage
         .get_complete_index_publication()
         .context("load complete core publication for lexical source policy")?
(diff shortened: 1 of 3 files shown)


TheGreenCedar:
Independent acceptance passed on `7aa80fc64f1ccddcb473343965076df255b07e05`, tree `32172aa68b595adff9cde3698aac8879c31d583c`. Sixteen full/descriptor observations preserve cache paths and bytes; fresh, warm and restarted ordinary managed search reach full Metal retrieval without creating the logical flat database. A damaged publication pointer fails closed without cache changes or native execution. The actual support archive executable SHA-256 is `ff60606d6d986e32afe557ef697e6ec2122ac766867e96665ac4b45b1b5cb121`; its native completion receipt records 108 tokens, MTL/MTL0 and CPU fallback disabled.

The mismatched-target descriptor fixture lacks lexical artifacts, so it does not independently prove identity rejection with otherwise healthy artifacts. Prior C2/089 failures remain retained. Required CI is pending; merge must match this head. Next: authenticate the final integration archive and rerun the same fresh installed interaction, then start the unchanged 36-session maintenance comparison. No release qualification or publication is claimed.

Issue #2210, opened 2026-09-14
Retire managed extraction before publishing runtime


TheGreenCedar:
A managed runtime becomes visible before its disposable extraction directory is retired. `provisionManagedCli` verifies managed staging, atomically renames it to the durable version directory, and rehashes the published binary through `resolveManifest`; extraction cleanup runs afterward in `finally`. A publisher exit in that interval can leave both a usable runtime and a duplicate extraction. Successful S06 development runs retain matching 224-MB extraction copies, but their creator/shutdown point is not established and is not attributed to this window.

First reproduce the publication-boundary failure against unchanged integrated source. Use the existing launcher fixtures and a deterministic test-only child interruption immediately after the actual managed-directory rename; avoid a large timing-dependent payload or any production delay/test hook. Verify the published manifest/runtime digest and subsequent reuse, and assert the extraction is absent (red on the old order).

If reproduced, retire only the current invocation's verified extraction after staged verification and before atomic publication, nulling its local pointer only after successful removal. Keep finally cleanup for failure paths. Preserve publication locking, version/stdio/digest checks, staging and rollback visibility, quarantine/download-cache behavior, source identity and native filesystem boundaries. No broad prefix reaper, cache eviction, API/schema or model/vector change.

Ownership: Astra/root owns Git, controls, integration and all execution. Sol/high owns one isolated source slice; Terra/xhigh owns independent exact-content challenge. Base `e61af7c5c69e2554549d3efcaa76b8fae45a4341`, tree `1b4c3f2380edd522e3c93f1e90cc69e38ffa401a`. Intended branch `codex/0176-install-temp`, worktree `/Users/albert/.codex/worktrees/0176-install-temp/CodeStory`. Allowed production file `plugins/codestory/scripts/codestory-mcp.cjs`; focused regression `plugins/codestory/tests/plugin-static.test.mjs`; user-visible Unreleased note if the fix is accepted. Tests wait for the active S06-Q15 native/resource run to close; no shared-temp test work underneath it.

Acceptance: the same before/after publication-boundary test, focused normal/failure/contended provisioning and download-retention contracts, plugin-static required lane, independent hostile/source review, applicable draft CI and exact integration identity. The sub-90-second native filesystem probe precedes any package rebuild for this slice. No S06 historical-cause, CPython completion, whole-host peak-fit or release claim follows from source proof. Prior issue #967/PR #980 and PR #930 contracts remain preserved.

Refs #2135, #2173, #2187

[...]

PR #2211, opened 2026-09-14
Retire verified extraction before runtime publication

TheGreenCedar:
A managed runtime could become visible before its temporary extraction was removed. If the host exited immediately after publication, the valid runtime was reused on restart while a duplicate extraction remained. A cleanup error could likewise arrive after the new runtime was already published.

This change removes verified extraction before the atomic publication rename. It binds cleanup to the created directory's native device/inode identity, rejects an observed replacement or symlink, and releases the managed lock even when cleanup fails. Existing staging verification and atomic publication remain in place. The identity check detects replacement before cleanup; it does not make the subsequent filesystem operation immune to a hostile same-account race.

Verification on `81ec20c594211d55df6e01ba14b8f92fd57e5399` (tree `f55935fb640f7504bf5821a924de57fa86205df2`):
- Deterministic old-source failures reproduced post-publication retention and cleanup-after-publication. Both pass after the change.
- Independent review's directory-replacement counterexample reproduced deletion and publication on the first revision. The revised regression proves replacement and prior-runtime preservation, no new publication, and lock release.
- Plugin static suite: 145 passed, one Windows-native locking case skipped. Installer suite: six passed. Diff and documentation-link checks pass.
- Independent Terra review accepted the bounded source repair; exact-head receipt and draft CI remain attached to this lane before readiness.

This is an unpublished development slice for unified 0.17.6. It does not attribute the historical retained S06 directories to this exit window, establish Windows-native behavior, or prove original CPython preparation fits its frozen budget. Candidate26 remains immutable during the current development investigation; final qualification follows product convergence.

Closes #2210
Refs #2135, #2173, #2187


TheGreenCedar:
Independent source acceptance is complete at `81ec20c594211d55df6e01ba14b8f92fd57e5399`, tree `f55935fb640f7504bf5821a924de57fa86205df2`. Three source files are clean and pushed. Root executed the independent directory-replacement counterexample red then green; final plugin checks passed 145/146 with one Windows-native skip, installer checks passed 6/6, and diff/doc-link checks passed.

The cleanup identity guard has a bounded claim: it rejects replacement observed before deletion and releases the lock on cleanup failure. It does not claim race-free deletion against a hostile same-account actor or Windows-native proof. Initial failure, rejected revision and all receipts remain preserved.

PR stays draft for applicable CI. Integration is additionally queued until the admitted S06-Q20 candidate26 investigation completes and native/dev ownership is released. No main promotion or final qualification is authorized by this receipt.

